PR #1, opened 2021-07-20
update deps + stop wasting cpu cycles + move main to examples

shikoku1:


shikoku1:
Non mi fa mettere @antonio gisondi

PR #5, opened 2021-08-02
Implement sdk

shikoku1:


src/lib.rs
```diff
@@ -1,215 +1,541 @@
...
-pub struct DeviceBuilder {
+/// Builder for Astarte client
+///
+/// ```
+/// use astarte_sdk::AstarteOptions;
+///
+/// let realm = "test";
+/// let device_id = "2TBn-jNESuuHamE2Zo1anA";
+/// let credentials_secret = "G6PjaFRwhjKLb07hk/lAUHqFoI+aIg39CEQ1MQ+mdZk=";
+/// let pairing_url = "https://api.eu1.astarte.cloud/pairing";
```
harlem88, line 57: Remove explicit Astarte info from comments

harlem88 (changes requested):
Remove explicit Astarte info from comments

src/lib.rs
Kezii: You mean the pairing url? The rest of the variables are all fake data

shikoku1: You mean the pairing url? The rest of the variables are all fake data

harlem88 (approved):

examples/simple.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    sdk_options
+        .add_interface_files(&interfaces_directory)
+        .unwrap();
+
+    sdk_options.build().await.unwrap();
+
+    let mut device = sdk_options.connect().await.unwrap();
+
+    let w = device.clone();
```
rbino, line 46: What happens to the MQTT client here? Will there be two separate clients with two connections after this call?

rbino (changes requested):

examples/simple.rs
shikoku1: immagine
There should be only one client with one connection

examples/simple.rs
```diff
@@ -0,0 +1,74 @@
...
+    tokio::task::spawn(async move {
+        let mut i: f64 = 0.0;
+        loop {
+            w.send("com.test.everything", "/double", i).await.unwrap();
```
bettio, line 50: let's use same naming conventions everywhere: `"com.test.everything"` -> "com.test.Everything"

bettio: Also we might use https://raw.githubusercontent.com/astarte-platform/astarte/master/standard-interfaces/org.astarte-platform.genericsensors.Values.json that is always good for our examples

src/interface.rs
```diff
@@ -351,7 +364,7 @@ mod tests {
 
         let interface = Interface::Datastream(datastream_interface);
 
-        let deser_interface = Interface::from_str(interface_json).unwrap();
+        let deser_interface = Interface::from_str(interface_json).unwrap(); //allow_panic
```
bettio, line 367: what does it mean this comment?

examples/datatypes.rs
```diff
@@ -0,0 +1,120 @@
...
+        "datetimearray",
+    ];
+
+    tokio::task::spawn(async move {
+        loop {
+            let data = alltypes.iter().zip(allendpoints.iter());
+
+            // individual aggregation
+            for i in data {
+                w.send("com.test.everything", &format!("/{}", i.1), i.0.clone())
```
bettio, line 94: I think we should use standard interfaces also here:
https://raw.githubusercontent.com/astarte-platform/astarte/master/standard-interfaces/org.astarte-platform.genericsensors.Values.json

examples/datatypes.rs
```diff
@@ -0,0 +1,120 @@
...
+                std::thread::sleep(std::time::Duration::from_millis(5));
+            }
+
+            // object aggregation
+            let mut obj: std::collections::HashMap<&str, astarte_sdk::types::AstarteType> =
+                std::collections::HashMap::new();
+            obj.insert("bottone", true.into());
+            obj.insert("uptimeSeconds", 67.into());
+
+            w.send_object_timestamp("com.test4.object", "/", obj, None)
```
bettio, line 107: I would use the GPS interface as example

bettio: @rbino any opinion about using standard interfaces ?

src/lib.rs
```diff
@@ -1,225 +1,675 @@
...
 
-    async fn populate_credentials(&mut self) -> Result<(), PairingError> {
-        let cert_pem = pairing::fetch_credentials(&self).await?;
-        let mut cert_pem_bytes = cert_pem.as_bytes();
-        let certs = pemfile::certs(&mut cert_pem_bytes).unwrap();
-        self.certificate_pem = Some(certs);
-        Ok(())
+    /// Serialize data directly from Bson
+    fn serialize(
+        data: Bson,
```
bettio, line 532: an error should be reported when data is an invalid type, or an invalid value (e.g. float NaN, ovf, etc...).

bettio (changes requested):

examples/simple.rs
shikoku1: I'd like to keep com.test.everything because it tests all the data types, I can remove the other one

src/interface.rs
shikoku1: I have a script that checks and warns for all the unwrap and panic, I'll remove the comments if you want

src/lib.rs
shikoku1: Should it be reported when sending? Or during serialization? On while constructing AstarteType?

shikoku1: I'd like to do this on the next pr

rbino (approved):

shikoku1:
Since this is not getting merged, I'm converting to draft and I'll merge here the next stuff, I'll remove the draft when this second patchset is finished

src/database.rs
```diff
@@ -0,0 +1,186 @@
...
+    async fn load_prop(&self, key: &str, version: i32)
+        -> Result<Option<AstarteType>, AstarteError>;
+    async fn delete_prop(&self, key: &str) -> Result<(), AstarteError>;
+
+    async fn clear(&self) -> Result<(), AstarteError>;
+}
+
+#[async_trait]
+impl AstarteDatabase for Database {
+    async fn store_prop(&self, key: &str, value: &[u8], version: i32) -> Result<(), AstarteError> {
```
rbino, line 26: Use `interface_major` instead of `version` both here and in the column, otherwise it's not clear what the version is referring to

rbino (changes requested):

rbino (approved):

bettio (approved):

PR #8, opened 2021-09-27
Refactor send object api

shikoku1:
hashmap was bad so we use a struct


harlem88 (approved):

rbino (approved):

rbino (changes requested):
Actually @shikoku1 those two commits can be squashed too, please do so

rbino (approved):

bettio (approved):

PR #9, opened 2021-10-04
Database state for sending properties 

shikoku1:


src/lib.rs
```diff
@@ -624,15 +389,25 @@ impl AstarteSdk {
...
         }
 
+        if self
+            .check_property_on_send(interface_name, interface_path, data.clone())
+            .await?
+        {
+            debug!("property was already sent, no need to send it again");
+            return Ok(());
+        }
+
```
rbino, line 410: There should be this kind of check also for unset (to avoid resending the unset)

shikoku1:
#10

PR #14, opened 2021-11-23
Update bson dependency

shikoku1:
We have to specialize the `from_bson_array` macro for `DateTimeArray` because bson has its own datetime type that doesn't support `FromIterator`


harlem88 (approved):

rbino (changes requested):
Please add to the commit message and PR description the implications of the update (e.g. the fact that you had to add some specialization for the DateTimeArray type).

shikoku1:
It's already documented in the code, there are no implications for the users

rbino (approved):

PR #19, opened 2021-11-24
Refactor and fix deserialization for unset

shikoku1:
The previous implementation was inconsistent and broke deserialization


rbino (changes requested):
Is there a reason for this change? If so please document it in the commit message so we also keep track of _why_ this was done

rbino (approved):

PR #27, opened 2021-12-14
Protocol: fix behavior after device `emptyCache` message

Annopaolo:
Do not clear the properties database after sending an  `emptyCache` message. This is a bug which may lead to device information loss. Instead, send all device-owned properties, as in e.g. the [Java](https://github.com/astarte-platform/astarte-device-sdk-java/blob/3de54b55d6074d62e468c2cef7030f791426fae3/DeviceSDK/src/main/java/org/astarteplatform/devicesdk/transport/mqtt/AstarteMqttV1Transport.java#L466-L474) or [Go](https://github.com/astarte-platform/astarte-device-sdk-go/blob/bf45b7921f32edf371a85f5be5a2bfcf3d6b515c/device/protocol_mqtt_v1.go#L187-L233) SDKs. 

In order to clean (parts) of the properties database, handling of purge properties messages should be implemented, as stated in the [Astarte doc](https://docs.astarte-platform.org/latest/080-mqtt-v1-protocol.html#purge-properties).

### DIFF
--- a/src/database.rs
+++ b/src/database.rs
@@ -21,6 +21,7 @@ use std::str::FromStr;
 
 use log::{debug, trace};
 use sqlx::sqlite::{SqliteConnectOptions, SqlitePoolOptions};
+use sqlx::FromRow;
 
 use crate::{types::AstarteType, AstarteError, AstarteSdk};
 
@@ -28,6 +29,15 @@ use crate::{types::AstarteType, AstarteError, AstarteSdk};
 #[derive(Clone, Debug)]
 pub struct AstarteSqliteDatabase {
     db_conn: sqlx::Pool<sqlx::Sqlite>,
+}
+
+/// This struct represents a property stored in the database
+#[derive(FromRow, Debug, PartialEq)]
+pub struct StoredProp {
+    pub interface: String,
+    pub path: String,
+    pub value: Vec<u8>,
+    pub interface_major: i32,
 }
 
 /// Database backend for the astarte client can be made by implementing this trait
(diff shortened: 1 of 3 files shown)


src/database.rs
```diff
@@ -50,6 +50,9 @@ pub trait AstarteDatabase {
 
     /// Removes all saved properties from the database
     async fn clear(&self) -> Result<(), AstarteError>;
+
+    /// Retrieves all property values in the database, together with their interface name, path and major version
+    async fn load_all_props(&self) -> Result<Vec<(String, String, Vec<u8>, i32)>, AstarteError>;
```
shikoku1, line 55: I think it's better to make a struct instead of this giant tuple
```rust
struct DbProp{
        interface: String,
        path: String,
        value: Vec<u8>,
        interface_major: i32,
}
...

async fn load_all_props(&self) -> Result<Vec<DbProp>, AstarteError>;
```

The struct is also useful for querying: 
```rust
sqlx::query_as!(DbProp, "select ...
```

shikoku1 (changes requested):

src/database.rs
```diff
@@ -220,5 +241,10 @@ mod test {
         db.clear().await.unwrap();
 
         assert_eq!(db.load_prop("com.test", "/test", 1).await.unwrap(), None);
+
+        // load all props
+        db.store_prop("com.test", "/test", &ser, 1).await.unwrap();
+        db.store_prop("com.test2", "/test", &ser, 1).await.unwrap();
+        assert_eq!(db.load_all_props().await.unwrap().len(), 2);
```
shikoku1, line 248: I'd test the correctness of the result too, for example, store a single property, call load_all_props to get it and then check every field for correctness

shikoku1 (approved):

rbino (approved):

harlem88 (approved):

PR #29, opened 2021-12-22
Add purge properties logic

shikoku1:
https://docs.astarte-platform.org/latest/080-mqtt-v1-protocol.html#purge-properties

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -154,12 +154,12 @@ impl AstarteSdk {
                             let topic = parse_topic(&p.topic);
 
                             if let Some((_, _, interface, path)) = topic {
+                                let bdata = p.payload.to_vec();
+
                                 if interface == "control" && path == "/consumer/properties" {
-                                    // TODO: implement consumer purge properties
+                                    self.purge_properties(bdata).await?;
                                     continue;
                                 }
-
-                                let bdata = p.payload.to_vec();
 
                                 debug!("Incoming publish = {} {:?}", p.topic, bdata);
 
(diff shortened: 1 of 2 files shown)


harlem88 (approved):

src/lib.rs
```diff
@@ -655,6 +674,22 @@ impl fmt::Debug for AstarteSdk {
...
 
+mod utils {
+    pub fn extract_purgeproperties(bdata: &[u8]) -> Vec<String> {
+        use flate2::read::ZlibDecoder;
+        use std::io::prelude::*;
+
+        let mut bdata = bdata.to_owned();
+        bdata.drain(0..4);
+
+        let mut d = ZlibDecoder::new(bdata.as_slice());
```
rbino, line 685: Are these 3 lines equivalent to
```
let mut d = ZlibDecoder::new(&bdata[4..]);
```
or am I missing something?

src/lib.rs
```diff
@@ -655,6 +674,22 @@ impl fmt::Debug for AstarteSdk {
     }
 }
 
+mod utils {
+    pub fn extract_purgeproperties(bdata: &[u8]) -> Vec<String> {
```
rbino, line 678: I'd call this `extract_set_properties` since the returned properties here are those that are _not_ purged (and it's more readable with some kind of whitespace between words). `set` can also be `kept`, `valid`, `existing`, whatever sounds better to you to convey the concept.

rbino (changes requested):

src/lib.rs
shikoku1: yes

rbino (approved):

Issue #20, opened 2021-11-25
Meta: Wish Support bitbake


rzr:
Hi,
Let me open issue to track related effort to support bitbake,
I have generated a recipe I will share it if you want to test until it's finalized.

Relate-to: https://github.com/astarte-platform/astarte-device-sdk-rust/pull/16

rzr:
Hi,
I have shared a recipe above, but it depends on other changes I am currently upstreaming:

Origin: https://github.com/openembedded/openembedded-core/commit/6cd2cf6525bcb241b3a2538e559fcef2a2084a7e
Relate-to: https://github.com/libunwind/libunwind/pull/312
Relate-to: https://booting.oniroproject.org/distro/oniro/-/issues/191
Forwarded: https://lists.openembedded.org/g/openembedded-core/message/158932

Fell free to reach me if you want to know more about status, some help will be also needed to use the cloud example.

I am not sure I used the right variables in:

```
RUST_BACKTRACE=1 cargo run --example simple -- --credentials-secret $credentialssecret --device-id $deviceid --pairing-url $pairingurl --realm $realm
```

rzr:
After configured interfaces at cloud side , i am able to see some activities:

![astarte-cloud](https://user-images.githubusercontent.com/89133/149620560-f164a0c2-11f3-4cbf-9f64-fc863005807b.png)

Next, it would be nice to make a shared lib that can be used from non rust programs too, what do you think of this idea ?

PR #31, opened 2022-01-15
docker: Add helper recipe

rzr:
My distro RUST setup is not running it out of the box
so I used a more recent one.

For the record here is is the observed error:

  error[E0599]: no variant or associated item named `OutOfMemory` found for enum `ErrorKind` in the current scope
  -> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/sqlx-core-0.5.10/src/sqlite/connection/establish.rs:107:32

Relate-to: https://github.com/astarte-platform/astarte-device-sdk-rust/issues/20#
Signed-off-by: Philippe Coval <philippe.coval@astrolabe.coop>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,38 @@
+#!/bin/echo docker build . -f
+# -*- coding: utf-8 -*-
+# Copyright 2022 Huawei OTC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+FROM rust:bullseye
+LABEL maintainer="Philippe Coval (philippe.coval@astrolabe.coop)"
+
+RUN echo "#log: Setup system" \
+  && set -x \
+  && apt-get update \
+  && apt-get install -y \
+      --no-install-recommends \
+     ca-certificates \
+     libssl-dev \
+     pkg-config \
+  && apt-get upgrade -y \
+  && apt-get clean \
+  && rm -rf /var/lib/apt/lists/* \
+  && date -u
+
+ARG project=astarte-device-sdk-rust
+ARG workdir="/local/src/${project}"
+
+COPY . "${workdir}/"
+WORKDIR "${workdir}"
+RUN echo "#log: ${project}: Building sources" \
+  && set -x \
+  && export OPENSSL_NO_VENDOR=1 \
+  && cargo build  \
+  && cargo install --path . || echo "warning: no default install rule" \
+  && cargo install --examples --path . \
+  && date -u
+
+WORKDIR "${workdir}"
+ENTRYPOINT [ "cargo" ]
+CMD [ "run", "--example", "simple", "--" , "--help" ]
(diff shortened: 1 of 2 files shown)


shikoku1:
> My distro RUST setup is not running it out of the box

For development you should consider using rustup instead of the distribution's packaged compiler, [as recommended by the rust project](https://www.rust-lang.org/tools/install)
Similarly, it might be better for the container to be based on the official rust image https://hub.docker.com/_/rust

Also, the docker build should be added to the CI if we want to support this

rzr:
ok I thought about this too

rzr:
"the docker build should be added to the CI if we want to support this" this can be done in next patch using github actions ?

Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+ARG workdir="/local/src/${project}"
+
+COPY . "${workdir}/"
+WORKDIR "${workdir}"
+RUN echo "#log: ${project}: Building sources" \
+  && set -x \
+  && export OPENSSL_NO_VENDOR=1 \
+  && cargo build  \
+  && cargo install --path . \
+  || cargo install --examples --path . \
```
shikoku1, line 38: is this really needed?

Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+  && date -u
+
+WORKDIR "${workdir}"
+ENTRYPOINT [ "cargo" ]
+CMD [ "run", "--example", "simple", "--help" ]
```
shikoku1, line 43: I'm not sure what the expected usage is, feel free to add instructions into the readme

Dockerfile
rzr: let me double check

Dockerfile
rzr: this is the default run command , it could be overridden when using docker run ...

shikoku1: Yeah but now it's calling --help for cargo, overriding --example, it doesn't do anything as it is. I can get what the idea was but I think we should fix this and document the intended usage anyway, because a dockerfile can be use in different ways (for example to produce binaries in a volume)

Dockerfile
```diff
@@ -0,0 +1,43 @@
+#!/bin/echo docker build . -f
+# -*- coding: utf-8 -*-
+# SPDX-License-Identifier: MPL-2.0
```
shikoku1, line 3: The rest of the project is licensed under Apache 2.0, is it ok to have a single file under a different license?
CC: @bettio

Dockerfile
rzr: yes it's needed I have updated the commit message

Dockerfile
rzr: --help for example will be displayed not cargo

Dockerfile
rzr: mistake let me align

PR #32, opened 2022-01-20
Add end to end tests

### LINKED ISSUE #30: Add workflow for automated testing using a Astarte Cluster action
See also: https://github.com/marketplace/actions/astarte-local-cluster

shikoku1:
Fix #30


.github/workflows/e2e-test.yaml
```diff
@@ -0,0 +1,38 @@
...
+          echo "RUST_LOG=astarte_sdk=debug" >> $GITHUB_ENV
+
+      - name: Install toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+      - name: Run test
+        run: bash e2etest.sh
```
harlem88, line 38: Missing new line at the end of file

e2etest.sh
```diff
@@ -0,0 +1,18 @@
...
+
+
+cargo run --example e2etest
+
+if [ $? -eq 0 ];then
+exit 0
+fi
+echo "Test failed, retrying"
+
+cargo run --example e2etest
```
harlem88, line 18: Missing new line at the end of file

Annopaolo, line 18: Why is the same test run three times?

examples/e2etest.rs
```diff
@@ -0,0 +1,70 @@
...
+            if let Err(err) = ret {
+                error!("send error {:?}", err);
+                std::process::exit(1);
+            }
+
+            std::thread::sleep(std::time::Duration::from_millis(300));
+        }
+
+        std::thread::sleep(std::time::Duration::from_millis(1000));
+        println!("Test completed successfully");
```
Annopaolo, line 55: For the e2e test to be meaningful, one should also check that the value is correctly received from Astarte. This is important especially for datastream interfaces having  mappings with reliability `unreliable` (default). At the moment, we do not know if data have actually reached Astarte.

To retrieve data, you can use any HTTP client:
 - Generate the JWT for authorization using astartectl: `astartectl utils gen-jwt all-realm-apis`. 
 - Query AppEngine at `/{realm_name}/devices/{device_id}/interfaces/{interface}[/{path}]`, 
 - Parse the resulting JSON object and finally compare with expected values.

.github/workflows/e2e-test.yaml
```diff
@@ -0,0 +1,38 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create Astarte Cluster
+        id: astarte
+        uses: astarte-platform/astarte-cluster-action@v1
+      - name: Checkout sources
+        uses: actions/checkout@v2
+      - name: Install interface
+        run: | 
+          astartectl realm-management interfaces install $GITHUB_WORKSPACE/examples/interfaces/com.test.Everything.json
```
Annopaolo, line 21: For tests, it is better to use the "canonical" prefix for interface names: `org.astarte-platform`...

examples/e2etest.rs
```diff
@@ -0,0 +1,70 @@
...
+        .unwrap()
+        .ignore_ssl_errors()
+        .build();
+
+    let mut device = astarte_sdk::AstarteSdk::new(&sdk_options).await.unwrap();
+
+    let w = device.clone();
+    tokio::task::spawn(async move {
+        for i in 0..3 {
+            let ret = w.send("com.test.Everything", "/integer", i).await;
```
Annopaolo, line 44: I'd say that testing only one endpoint can be too little, but this can be tackled on latter iterations.

Annopaolo (changes requested):
Some more to be done, but the PR is good. 
It is better to move e2e test files to a separate directory, as done [here](https://github.com/astarte-platform/astarte/tree/master/tools/astarte_e2e) for example.

e2etest.sh
shikoku1: this should no longer be necessary after #35

examples/e2etest.rs
shikoku1: Since the other pr is temporarily blocked, I have made another commit here with those checks, it is now ready for review (again)

.github/workflows/e2e-test.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+on:
+  push:
+  # Run on branch/tag creation
+  create:
+  # Run on pull requests
+  pull_request:
+
+jobs:
+  create-cluster:
```
Annopaolo, line 29:
```suggestion
  e2e-test: 
```

examples/e2etest.rs
```diff
@@ -0,0 +1,384 @@
+/*
```
Annopaolo, line 1: Examples should be separated from e2e tests, i suggest you use another directory as in other Astarte repos

Annopaolo (changes requested):
A couple minor changes, but the PR is ok.

harlem88 (approved):

.github/workflows/e2e-test.yaml
```diff
@@ -0,0 +1,59 @@
...
+    steps:
+      - name: Create Astarte Cluster
+        id: astarte
+        uses: astarte-platform/astarte-cluster-action@v1
+      - name: Checkout sources
+        uses: actions/checkout@v2
+      - name: Install interface
+        run: | 
+          astartectl realm-management interfaces install $GITHUB_WORKSPACE/examples/interfaces/org.astarte-platform.test.Everything.json
+          astartectl realm-management interfaces install $GITHUB_WORKSPACE/examples/interfaces/org.astarte-platform.genericsensors.Geolocation.json
```
Annopaolo, line 40:
```suggestion
          astartectl realm-management interfaces sync $GITHUB_WORKSPACE/examples/interfaces/*.json --non-interactive
```

src/e2etest/e2etest.rs
```diff
@@ -0,0 +1,473 @@
...
+
+fn get_data_obj() -> HashMap<String, f64> {
+    let mut data: HashMap<String, f64> = HashMap::new();
+    data.insert("latitude".into(), 1.34);
+    data.insert("longitude".into(), 2.34);
+    data.insert("altitude".into(), 3.34);
+    data.insert("accuracy".into(), 4.34);
+    data.insert("altitudeAccuracy".into(), 5.34);
+    data.insert("heading".into(), 6.34);
+    data.insert("speed".into(), 7.34);
```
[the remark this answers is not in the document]
shikoku1: I originally wanted to do that but there's an issue with AstarteType serialization that blocks it, and I'm not really sure if it's worth to do a big internal change to allow this if there are no production requirements

bettio (approved):

PR #33, opened 2022-01-21
Update interfaces with the client already connected

shikoku1:
Add method to update interfaces with the client already open
A refactoring was needed to add interior mutability to the astarte sdk struct, now the struct AstarteOptions is saved inside AstarteSdk

### DIFF
--- a/src/interfaces.rs
+++ b/src/interfaces.rs
@@ -20,9 +20,11 @@
 
 use std::collections::HashMap;
 
+use log::debug;
+
 use crate::{interface::traits::Mapping, types::AstarteType, AstarteError, Interface};
 
-#[derive(Clone)]
+#[derive(Clone, Default)]
 pub struct Interfaces {
     pub interfaces: HashMap<String, Interface>,
 }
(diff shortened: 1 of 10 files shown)


src/interfaces.rs
```diff
@@ -276,21 +278,51 @@ impl Interfaces {
...
+        interfaces_directory: &str,
+    ) -> Result<(), AstarteError> {
+        let interface_files = std::fs::read_dir(std::path::Path::new(interfaces_directory))?;
+        let it = interface_files.filter_map(Result::ok).filter(|f| {
+            if let Some(ext) = f.path().extension() {
+                ext == "json"
+            } else {
+                false
+            }
+        });
```
harlem88, line 303: How about using this here? `.for_each(move |f| self.add_interface_file(&f.path().to_string_lossy())?);`

shikoku1: can't use the question mark operator from inside a closure

src/interfaces.rs
```diff
@@ -276,21 +278,51 @@ impl Interfaces {
...
+    }
+
+    /// Add all json interface description inside a specified directory
+    pub fn add_interface_directory(
+        &mut self,
+        interfaces_directory: &str,
+    ) -> Result<(), AstarteError> {
+        let interface_files = std::fs::read_dir(std::path::Path::new(interfaces_directory))?;
+        let it = interface_files.filter_map(Result::ok).filter(|f| {
+            if let Some(ext) = f.path().extension() {
```
harlem88, line 298: How about using `ends_with` to check extension ?
```suggestion
            .filter(|f| f.path().ends_with("json"));
```

harlem88 (changes requested):

harlem88 (approved):

src/lib.rs
```diff
@@ -174,13 +192,24 @@ impl AstarteSdk {
...
     }
 
+    pub async fn update_interfaces(
+        &self,
+        interfaces: interfaces::Interfaces,
+    ) -> Result<(), AstarteError> {
+        self.opts.write().await.interfaces = interfaces;
+
+        self.subscribe().await?;
+        self.send_introspection().await?;
```
Annopaolo, line 202: When introspection update at runtime is allowed, some annoying corner cases appear. The solution used is a number of SDKs is to persist the current introspection and save the new one only when it's acknowledged correctly by Astarte.
You can take a look at the [Qt5 SDK](https://github.com/astarte-platform/astarte-device-sdk-qt5/commit/a521af3e1160bc6ad6aeb43a99b4e99473bb042f) or the [Go SDK](https://github.com/astarte-platform/astarte-device-sdk-go/pull/33) for this.

Annopaolo (changes requested):

src/lib.rs
shikoku1: this is blocked by this https://github.com/bytebeamio/rumqtt/issues/349 temporarily

PR #51, opened 2022-05-26
Fix handling of int32 payload on LongInteger mapping

FrancescoVaiani:
In the case of a server-owned longinteger mapping, Astarte sends an integer number if the value is small enough.
This behavior was not taken into account during the data validation. 
A condition was add in the `eq` function that compares the type of the data received with the mapping type.
Added some tests.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -22,6 +22,7 @@ use std::convert::TryInto;
 
 use bson::{Binary, Bson};
 
+use crate::interface::MappingType;
 use crate::AstarteError;
 
 /// Types supported by astarte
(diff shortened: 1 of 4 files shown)


src/types.rs
```diff
@@ -64,6 +64,12 @@ impl PartialEq<crate::interface::MappingType> for AstarteType {
             };
         }
 
+        if other == &crate::interface::MappingType::LongInteger {
```
harlem88, line 67: Specify at top of function the `use` declaration for better readability.
``` Rust
use crate::interface::MappingType 
``` 
....
``` Rust
if other == &MappingType::LongInteger
```

harlem88 (changes requested):

harlem88 (approved):

Annopaolo (approved):

PR #65, opened 2022-08-08
Add Interface after device connection

FrancescoVaiani:
Add the capability to dynamically manipulate the introspection adding an interface after the device connection.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -40,9 +40,12 @@ use rumqttc::{EventLoop, MqttOptions};
 use std::collections::HashMap;
 use std::convert::TryInto;
 use std::fmt::{self, Debug};
+use std::path::Path;
 use std::sync::Arc;
 use types::AstarteType;
 
+use crate::builder::AstarteBuilderError;
+use crate::interface::traits::Interface as iface;
 pub use interface::Interface;
 
 /// Astarte client
@@ -53,7 +56,7 @@ pub struct AstarteSdk {
     mqtt_options: MqttOptions,
     client: AsyncClient,
     eventloop: Arc<tokio::sync::Mutex<EventLoop>>,
-    interfaces: interfaces::Interfaces,
+    interfaces: Arc<tokio::sync::RwLock<interfaces::Interfaces>>,
     database: Option<Arc<dyn AstarteDatabase + Sync + Send>>,
 }
 
@@ -138,7 +141,9 @@ impl AstarteSdk {
             mqtt_options,
             client,
             eventloop: Arc::new(tokio::sync::Mutex::new(eventloop)),
-            interfaces: interfaces::Interfaces::new(opts.interfaces.clone()),
+            interfaces: Arc::new(tokio::sync::RwLock::new(interfaces::Interfaces::new(
+                opts.interfaces.clone(),
+            ))),
             database: opts.database.clone(),
         };
 
@@ -148,10 +153,8 @@ impl AstarteSdk {
     }
 
     async fn subscribe(&self) -> Result<(), AstarteError> {
-        let ifaces = self
-            .interfaces
-            .interfaces
-            .clone()
+        let ifaces = &self.interfaces.read().await.interfaces;
+        let server_owned_ifaces = ifaces
             .into_iter()
             .filter(|i| i.1.get_ownership() == interface::Ownership::Server);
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -339,7 +381,9 @@ impl AstarteSdk {
     }
 
     async fn send_introspection(&self) -> Result<(), AstarteError> {
-        let introspection = self.interfaces.get_introspection_string();
+        let interfaces_arc = self.interfaces.clone();
```
harlem88, line 384: Call this `interfaces_mutex` as all other cases where you cloned the `self.interfaces`.

src/lib.rs
```diff
@@ -599,8 +665,10 @@ impl AstarteSdk {
 
             let data: AstarteType = data.into();
 
-            let mapping = self
-                .interfaces
+            let interfaces_mutex = self.interfaces.clone();
+
```
harlem88, line 669: Remove new line.

src/lib.rs
```diff
@@ -636,8 +704,10 @@ impl AstarteSdk {
...
```
harlem88, line 708: Same above.

harlem88 (changes requested):

harlem88 (approved):

src/lib.rs
```diff
@@ -53,7 +56,7 @@ pub struct AstarteSdk {
     mqtt_options: MqttOptions,
     client: AsyncClient,
     eventloop: Arc<tokio::sync::Mutex<EventLoop>>,
-    interfaces: interfaces::Interfaces,
+    interfaces: Arc<tokio::sync::Mutex<interfaces::Interfaces>>,
```
rbino, line 59: Why are we using `tokio::sync::Mutex` here? [This section](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use) seems to imply that we would be better off using the one in the standard library or in `parking_lot` (that we already have as indirect dependency).
Moreover, can't we use a `RwLock` instead of a Mutex? I don't think we need to lock it every time we read from it, only when we write to it

rbino (changes requested):

src/lib.rs
```diff
@@ -150,6 +155,8 @@ impl AstarteSdk {
     async fn subscribe(&self) -> Result<(), AstarteError> {
         let ifaces = self
             .interfaces
+            .read()
+            .await
             .interfaces
             .clone()
```
rbino, line 161: Why are interfaces being cloned here? (Was already there but can be fixed nevertheless) I expect that they just have to be read

src/lib.rs
```diff
@@ -206,6 +221,28 @@ impl AstarteSdk {
...
+    pub async fn add_interface(&self, file_path: &str) -> Result<(), AstarteError> {
+        let path = Path::new(file_path);
+        match Interface::from_file(path) {
+            Ok(interface) => {
+                let interfaces_mutex = self.interfaces.clone();
+                let mut interfaces = interfaces_mutex.write().await;
+                let interfaces_map = &mut interfaces.interfaces;
+                if interface.get_ownership() == interface::Ownership::Server {
+                    self.subscribe_server_owned_interface(&interface).await?;
+                }
```
rbino, line 233: This can be moved to _before_ we acquire the lock since it doesn't need to be in the critical section

src/lib.rs
```diff
@@ -206,6 +221,28 @@ impl AstarteSdk {
         Ok(())
     }
 
+    pub async fn add_interface(&self, file_path: &str) -> Result<(), AstarteError> {
+        let path = Path::new(file_path);
+        match Interface::from_file(path) {
+            Ok(interface) => {
+                let interfaces_mutex = self.interfaces.clone();
```
rbino, line 228: Call this `lock` instead of `mutex` to reflect its actual type

rbino (changes requested):

src/lib.rs
FrancescoVaiani: We can avoid the `clone` borrowing `self.interfaces.interfaces`

src/lib.rs
```diff
@@ -148,10 +153,8 @@ impl AstarteSdk {
     }
 
     async fn subscribe(&self) -> Result<(), AstarteError> {
-        let ifaces = self
-            .interfaces
-            .interfaces
-            .clone()
+        let ifaces = &self.interfaces.read().await.interfaces;
+        let ifaces_filter = ifaces
```
rbino, line 157: Call this `server_owned_ifaces` so it's clearer what this represents

src/lib.rs
```diff
@@ -162,18 +165,26 @@ impl AstarteSdk {
...
+            self.subscribe_server_owned_interface(&iface).await?;
         }
 
         Ok(())
     }
 
+    async fn subscribe_server_owned_interface(
+        &self,
+        iface: &Interface,
+    ) -> Result<(), AstarteError> {
```
rbino, line 178: As it is this assumes the caller correctly passes just server owned interfaces, it would probably be a good idea to check it also here and either return an error if the interface is not server owned or print a warning and avoid subscribing to it

rbino (changes requested):

rbino (approved):

bettio (approved):

PR #71, opened 2022-09-08
API cleanup

### LINKED ISSUE #57: API Cleanup
- [ ] Rename `AstarteSdk` to `AstarteDeviceSdk`

FrancescoVaiani:
Rename package from `astarte_sdk` to `astarte-device-sdk` (automatically translated to `astarte_device_sdk` inside the code) to better adhere to common RUST naming conventions.
Rename various APIs for better understandability.

Close #57 

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,10 +6,9 @@
 
 
 [package]
-name = "astarte_sdk"
+name = "astarte-device-sdk"
 version = "0.1.0"
-authors = ["Riccardo Binetti"]
-edition = "2018"
+edition = "2021"
 license = "Apache-2.0"
 repository = "https://github.com/astarte-platform/astarte-device-sdk-rust"
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -50,7 +50,7 @@ pub use interface::Interface;
 
 /// Astarte client
 #[derive(Clone)]
-pub struct AstarteSdk {
+pub struct AstarteDeviceSdk {
     realm: String,
     device_id: String,
     mqtt_options: MqttOptions,
@@ -116,7 +116,7 @@ pub enum Aggregation {
 
 /// data from astarte to device
 #[derive(Debug, Clone)]
-pub struct Clientbound {
+pub struct AstarteDeviceDataEvent {
     pub interface: String,
     pub path: String,
     pub data: Aggregation,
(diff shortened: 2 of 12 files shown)


Cargo.toml
```diff
@@ -6,7 +6,7 @@
 
 
 [package]
-name = "astarte_sdk"
+name = "astarte-device-sdk"
 version = "0.1.0"
 authors = ["Riccardo Binetti"]
 edition = "2018"
```
harlem88, line 12: Change the edition to 2021.

harlem88 (changes requested):

src/lib.rs
```diff
@@ -210,18 +209,20 @@ impl AstarteSdk {
...
     ///     loop {
-    ///         if let Ok(data) = d.poll().await {
+    ///         if let Ok(data) = d.astarte_device_data_event().await {
     ///             println!("incoming: {:?}", data);
     ///         }
     ///     }
     /// }
     /// ```
-    pub async fn poll(&mut self) -> Result<Clientbound, AstarteError> {
+    pub async fn astarte_device_data_event(
```
harlem88, line 223: Maybe `astarte_device_event` is better because it also receives the `connack` and `publish` event. What do you think @Annopaolo ?

Annopaolo: Even if `connack` and `publish` are actually MQTT events, they play an important role in the Astarte MQTTv1 protocol. So I think it's correct to keep the `astarte_` prefix. I agree with your 'astarte_device_event` suggestion. Maybe also `astarte_device_event_handler` , but I'm not sure it conforms to Rust naming best practices.

FrancescoVaiani: I don't see any issue with the `astarte_device_event_handler` naming. I was also wondering about it myself.

harlem88 (approved):

Annopaolo (approved):

src/lib.rs
```diff
@@ -210,18 +209,20 @@ impl AstarteSdk {
...
     ///     loop {
-    ///         if let Ok(data) = d.poll().await {
+    ///         if let Ok(data) = d.astarte_device_event_handler().await {
     ///             println!("incoming: {:?}", data);
     ///         }
     ///     }
     /// }
     /// ```
-    pub async fn poll(&mut self) -> Result<Clientbound, AstarteError> {
+    pub async fn astarte_device_event_handler(
```
rbino, line 223: I'm not totally convinced by a couple of things:
- Do we need the `astarte_device` prefix? Methods are already namespaced in an `AstarteDeviceSdk` struct
- `event_handler` lets me think of a function _you provide_ to handle event, while here you have to _call_ the function to handle the events, so I'd probably call this `handle_events`, if we want to rename it. This is actually more akin to an eventloop. Alternative names that come to mind: keep `poll`, `tick`, `advance`, `feed`

src/lib.rs
```diff
@@ -174,7 +173,7 @@ impl AstarteSdk {
         Ok(())
     }
 
-    async fn poll_connack(&mut self) -> Result<(), AstarteError> {
+    async fn astarte_device_connack_event(&mut self) -> Result<(), AstarteError> {
```
rbino, line 176: See observations in the other comment for prefix and naming, I'd call this `wait_for_connack`

rbino (changes requested):

src/lib.rs
harlem88: vote for `handle_events`

sorru94 (approved):

harlem88 (approved):

PR #80, opened 2022-11-16
Add more ways to add interfaces to introspection

FrancescoVaiani:
- Alter `add_interface` to get an Interface as parameter. 
- Add `add_interface_from_file` and `add_interface_from_str` to add a serialized interfaces to the introspection.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -41,10 +41,10 @@ use std::collections::HashMap;
 use std::convert::TryInto;
 use std::fmt::{self, Debug};
 use std::path::Path;
+use std::str::FromStr;
 use std::sync::Arc;
 use types::AstarteType;
 
-use crate::builder::AstarteBuilderError;
 use crate::interface::traits::Interface as iface;
 pub use interface::Interface;
 
@@ -94,6 +94,9 @@ pub enum AstarteError {
 
     #[error("builder error")]
     BuilderError(#[from] builder::AstarteBuilderError),
+
+    #[error(transparent)]
+    InterfaceError(#[from] interface::Error),
 
     #[error("generic error")]
     Reported(String),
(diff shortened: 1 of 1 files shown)


src/interface.rs
```diff
@@ -224,6 +224,12 @@ impl Interface {
         Ok(interface)
     }
 
+    pub fn from_str(s: &str) -> Result<Self, Error> {
```
harlem88, line 227: Why redefine this function? It is already defined here https://github.com/astarte-platform/astarte-device-sdk-rust/blob/5d4ec0a5dc5736382e2f27d53fb1be1f2892f6f8/src/interface.rs#L317

src/lib.rs
```diff
@@ -221,24 +224,25 @@ impl AstarteSdk {
...
+    }
+
+    pub async fn add_interface(&self, interface: Interface) -> Result<(), AstarteError> {
+        if interface.get_ownership() == interface::Ownership::Server {
+            self.subscribe_server_owned_interface(&interface).await?;
         }
+        let interfaces_lock = self.interfaces.clone();
+        let mut interfaces = interfaces_lock.write().await;
+        let interfaces_map = &mut interfaces.interfaces;
+        interfaces_map.insert(interface.name().to_string(), interface);
```
Annopaolo, line 245: We should release the rwlock after this line, so that we do not wait for the message to be sent before allowing others (maybe even `send_introspection` itself, check [here](https://github.com/astarte-platform/astarte-device-sdk-rust/blob/5d4ec0a5dc5736382e2f27d53fb1be1f2892f6f8/src/lib.rs#L384)) to access `self.interfaces`.
It would also make possible to defer handling MQTT errors to a later time, using the Astarte introspection synchronization mechanism instead.

Annopaolo (changes requested):

harlem88 (approved):

Annopaolo (approved):

PR #81, opened 2022-11-16
Fix the magic number in `store_property_on_send`

FrancescoVaiani:
Replaced `0` magic number in `store_property_on_send` with the actual `version_major` meant to be there.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -746,14 +746,23 @@ impl AstarteSdk {
             let data: AstarteType = data.into();
 
             let interfaces = self.interfaces.read().await;
+            let interface_major = interfaces
+                .interfaces
+                .get(interface_name)
+                .ok_or_else(|| {
+                    AstarteError::InterfaceError(interface::Error::InterfaceNotFoundError)
+                })?
+                .get_version_major();
             let mapping = interfaces
                 .get_mapping(interface_name, interface_path)
-                .ok_or_else(|| AstarteError::SendError("Mapping doesn't exist".into()))?;
+                .ok_or_else(|| {
+                    AstarteError::InterfaceError(interface::Error::MappingNotFoundError)
+                })?;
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -698,14 +698,19 @@ impl AstarteSdk {
             let data: AstarteType = data.into();
 
             let interfaces = self.interfaces.read().await;
+            let interface_major = interfaces
+                .interfaces
+                .get(interface_name)
+                .ok_or_else(|| AstarteError::SendError("Interface doesn't exist".into()))?
```
Annopaolo, line 704: Isn't the existence of the interface verified earlier? E.g. [here](https://github.com/astarte-platform/astarte-device-sdk-rust/blob/5d4ec0a5dc5736382e2f27d53fb1be1f2892f6f8/src/lib.rs#L632-L634)

FrancescoVaiani: It's a safer alternative than `.unwrap()`, but I agree it's redundant.

Annopaolo (approved):

src/lib.rs
```diff
@@ -698,14 +698,19 @@ impl AstarteSdk {
             let data: AstarteType = data.into();
 
             let interfaces = self.interfaces.read().await;
+            let interface_major = interfaces
+                .interfaces
+                .get(interface_name)
+                .unwrap()
```
harlem88, line 704: Maybe is better to propagate the error with `?` than generate a panic in case of error or empty value.

harlem88 (changes requested):

src/lib.rs
harlem88: Since it is used in a separate function, perhaps it is better to keep a safer version.

harlem88 (approved):

harlem88 (approved):

PR #116, opened 2023-01-20
Handle the reception of an Integer value in a Mapping with type Double

### LINKED ISSUE #114: Receiving a Double that is actually an integer number
When receiving a Double that is actually an integer number (for example sending 42.0).  The device poll function returns the error: `ReceiveError("You are receiving the wrong type for this mapping").`

Additional info:
Astarte POST request body (for an individual datastream) is: {"data":42}.

harlem88:
The conversion from AstarteType Double or LongInteger supports also Integer, so the end user will be able to use the try_into trait, without having to match Integer/Double in separately block.

Add tests on the comparison between MappingType e AstarteType.

Close #114 .

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -65,7 +65,7 @@ impl PartialEq<MappingType> for AstarteType {
             };
         }
 
-        if other == &MappingType::LongInteger {
+        if other == &MappingType::LongInteger || other == &MappingType::Double {
             if let AstarteType::Integer(_) = self {
                 return true;
             }
@@ -319,6 +319,7 @@ mod test {
 
     use chrono::{DateTime, TimeZone, Utc};
 
+    use crate::interface::MappingType;
     use crate::{types::AstarteType, Aggregation, AstarteDeviceSdk, AstarteError};
 
     #[test]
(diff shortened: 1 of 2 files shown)


FrancescoVaiani:
Change the title to "Handle the reception of an Integer value in a Mapping with type Double".
In the description change ~`Added`~ with `Add`.

FrancescoVaiani (approved):

src/types.rs
```diff
@@ -65,7 +65,7 @@ impl PartialEq<MappingType> for AstarteType {
             };
         }
 
-        if other == &MappingType::LongInteger {
+        if other == &MappingType::LongInteger || other == &MappingType::Double {
```
sorru94, line 68: This fixes the problem, but there is a resulting behaviour that I would like to highlight.

For example. When receiving the data: `42` that is mapped in the interface as a `double` the `poll` function will return the value `AstarteType::Integer(42)`. 
The same problem was already present for the `longinteger` mapping. 
For example. When receiving the data `12` that is mapped in the interface as a `longinteger` the `poll` function will return the value `AstarteType::Integer(12)`.

For an interface of the type `double` the user should expect the `poll` function to produce an `AstarteType::Integer` or `AstarteType::Double`.

harlem88: Right, but in the case we have two options:
1. Because Astarte send the smallest type, also the SDK adapts to it.
2. The Astarte Device SDK should produce the AstarteType in according with the mapping of Interface, also because Rust doesn't allow to receive a different type.

sorru94: The second option could result in an easier to use API for a potential user.
With the current implementation each user needs an additional `match` block when receiving an endpoint that they expect to be a `Double` and then cast to an `f64`/`f32` separately for `Integer`/`Double`.
With option 2 they would just cast to `f64`/`f32` .

A third option might be to implement the `TryFrom` trait from `AstarteType::Integer` to `f32`/`f64`.

To be honest, regarding the first option I lack the knowledge to understand if that could be an ok behaviour or not. Not sure how the other SDK handle this, and how much it's a language related issue.

sorru94 (approved):

FrancescoVaiani (approved):

bettio (approved):

PR #120, opened 2023-01-24
Fix unset property from server

### LINKED ISSUE #110: Unset properties not functioning from Server
Unsetting properties for server owned interfaces does not correctly work.
When a server unsets a propery and sends the related message to the device, the device returns the following error.
```
poll error ReceiveError("You are receiving the wrong type for this mapping")
```
This error is triggered by the following check:
https://github.com/astarte-platform/astarte-device-sdk-rust/blob/8286c58a325ae0b4eddedb69c756e7a243ad7a77/src/interfaces.rs#L252-L256
As can be seen above the type of the received data is compared to the expected type from the interface mapping.
This creates an undesired behaviour when the received type is `AstarteType::Unset`.

I suggest the following fix:
```
if (individual != AstarteType::Unset) && (individual != mapping.mapping_type()) {
    return Err(AstarteError::ReceiveError(
    "You are receiving the wrong type for this mapping".into(),
    ));
}
```

This fix does not account if the property has the `allow_unset` flag set. Assuming the Server only sends unset messages for properties that are unsettable.

sorru94:
Closes #110.

### DIFF
--- a/src/interfaces.rs
+++ b/src/interfaces.rs
@@ -22,7 +22,7 @@ use std::collections::HashMap;
 
 use crate::{interface::traits::Mapping, types::AstarteType, AstarteError, Interface};
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct Interfaces {
     pub interfaces: HashMap<String, Interface>,
 }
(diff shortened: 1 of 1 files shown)


src/interfaces.rs
```diff
@@ -233,23 +233,19 @@ impl Interfaces {
...
+                            "Do not unset a mapping without allow_unset".into(),
+                        ));
                     }
                 }
 
-                if individual != mapping.mapping_type() {
+                if (individual != AstarteType::Unset) && (individual != mapping.mapping_type()) {
                     return Err(AstarteError::ReceiveError(
                         "You are receiving the wrong type for this mapping".into(),
                     ));
```
harlem88, line 251: What do you think if you rewrite something  like this?
``` Rust
if (individual == AstarteType::Unset) {
    if bdata.is_empty() && !map.allow_unset {
        return Err(AstarteError::ReceiveError( "Do not unset a mapping without allow_unset".into(),
        ));
    }
}
```
the validation of a property with `unset` fails only if  ` bdata.is_empty() && !map.allow_unset` , while is true in the other case, also the line 252 can remain unchanged.

harlem88 (changes requested):

src/interfaces.rs
sorru94: By reverting the guard: 
```
if (individual != AstarteType::Unset) && (individual != mapping.mapping_type()) {
```
to:
```
if individual != mapping.mapping_type() {
```
Unsetting a property would always fail as the mapping does not match.
Furthermore, the condition `bdata.is_empty()` is actually equivalent to  `individual == AstarteType::Unset`. As `individual` is set to `AstarteType::Unset` in the function `deserialize`.

harlem88 (approved):

Issue #70, opened 2022-09-02
Publishing to Crates.io


harlem88:
https://doc.rust-lang.org/book/ch14-02-publishing-to-crates-io.html

PR #128, opened 2023-01-27
Improved docstrings

### LINKED ISSUE #129: Lacking documentation for the crate
Before pubblishing on `crates.io` (#70) the documentation should be improved, documenting all public interfaces.

sorru94:
Adds docstrings to all the methods and data structures visible outside the crate.
Two other changes come bundled in this PR:
- Some of the methods of the `Interface` enumerator have been made public only inside of the crate. 
- The `handle_events()` has been splitted in two. This eases readability and prevents eccessive indentation.

Preparation for #70.
Closes #129.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,11 @@ SPDX-License-Identifier: Apache-2.0
 
 *Warning: this SDK is experimental, correctness and API stability are currently not guaranteed*
 
-This package allows you to implement an Astarte Device using Rust.
+The Astarte Device SDK for Rust is a ready to use library that provides communication and
+pairing primitives to an Astarte Cluster.
+
+See the [Astarte documentation](https://docs.astarte-platform.org/latest/001-intro_user.html)
+for more information regarding Astarte and the available SDKs.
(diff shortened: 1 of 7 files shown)


FrancescoVaiani (approved):

src/builder.rs
```diff
@@ -167,6 +176,7 @@ impl AstarteOptions {
         Ok(self)
     }
 
+    /// Clone the astarte options.
```
harlem88, line 179: Maybe is better something like `Returns a reference to the AstarteOptions being constructed.`

harlem88 (changes requested):

src/builder.rs
sorru94: True. 
By the way, what is the purpose of this method?
The name `build` seems to imply some complex operation, while it just clones the options.
Why not let the user take care of cloning?

harlem88 (approved):

FrancescoVaiani (approved):

PR #130, opened 2023-01-31
Remove AstarteOptions build method

FrancescoVaiani:
AstarteOptions does not follow the builder pattern, and a `build()` method that only clones the original object is superflous. Change the name and methods to better reflect the actual behavior of the AstarteOptions struct.

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -37,7 +37,7 @@ use crate::pairing;
 /// Builder for Astarte client
 ///
 /// ```no_run
-/// use astarte_device_sdk::builder::AstarteOptions;
+/// use astarte_device_sdk::options::AstarteOptions;
 ///
 /// #[tokio::main]
 /// async fn main(){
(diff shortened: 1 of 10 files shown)


src/options.rs
```diff
@@ -147,10 +140,10 @@ impl AstarteOptions {
     }
 
     /// Add all json interface description inside a specified directory
-    pub fn interface_directory<'a>(
-        &'a mut self,
+    pub fn interface_directory(
+        self,
```
sorru94, line 144:
```suggestion
        mut self,
```

src/options.rs
```diff
@@ -160,14 +153,12 @@ impl AstarteOptions {
             }
         });
 
+        let mut astarte_options = self;
```
sorru94, line 156:
```suggestion
```

src/options.rs
```diff
@@ -160,14 +153,12 @@ impl AstarteOptions {
...
+
         for f in it {
-            self.interface_file(&f.path())?;
+            astarte_options = astarte_options.interface_file(&f.path())?;
```
sorru94, line 159:
```suggestion
            self = self.interface_file(&f.path())?;
```

src/options.rs
```diff
@@ -160,14 +153,12 @@ impl AstarteOptions {
...
         }
 
-        Ok(self)
-    }
-
-    pub fn build(&self) -> Self {
-        self.clone()
+        Ok(astarte_options)
```
sorru94, line 162:
```suggestion
        Ok(self)
```

sorru94 (commented):
The four comments are for the same suggestion. Not sure if you like it more like this, but I think both should work.

harlem88 (approved):

sorru94 (approved):

PR #141, opened 2023-02-17
e2e tests: bump up version of Astarte Cluster action and fix a related issue.

### LINKED ISSUE #142: e2e tests: new version of Astarte action and fix related bug.
An incorrect check for the result of the `POST` and `DELETE` commands is present in the end-to-end tests. 
The error printed [here](https://github.com/astarte-platform/astarte-device-sdk-rust/actions/runs/4105730310/jobs/7082999083#step:7:403) should lead to a test failure.
Furthermore, the error mentioned above is a bug in Astarte that has been fixed from `v1.0.4`. The GitHub action for the Astarte cluster is by default stuck at `v1.0.1`. It should be forcefully updated to `v1.0.4` in order to avoid an incorrect test failure.

sorru94:
Some fixes for the end to end tests:
- Handle correctly a failure of `POST` or `DELETE` HTTPS requests.
- Bump up version of Astarte in CI to `1.0.4`
- Removes the not necessary `ignore_ssl_errors` flag in the device initialization.

Closes #142

### DIFF
--- a/tests/e2etest/main.rs
+++ b/tests/e2etest/main.rs
@@ -31,6 +31,7 @@ use std::time::Duration;
 use std::{env, panic, process};
 
 use colored::Colorize;
+use reqwest::StatusCode;
 use serde_json::Value;
 use tokio::{task, time};
 
@@ -118,6 +119,14 @@ impl TestCfg {
 
 #[tokio::main]
 async fn e2etest_impl() {
+    // Set custom panic hook to exit in case a subprocess panics.
+    let orig_hook = panic::take_hook();
+    panic::set_hook(Box::new(move |panic_info| {
+        // invoke the default handler and exit the process
+        orig_hook(panic_info);
+        process::exit(1);
+    }));
+
     let test_cfg = TestCfg::init().expect("Failed configuration initialization");
 
     let sdk_options = AstarteOptions::new(
@@ -127,8 +136,7 @@ async fn e2etest_impl() {
         &test_cfg.pairing_url,
     )
     .interface_directory(&test_cfg.interfaces_fld.to_string_lossy())
-    .unwrap()
-    .ignore_ssl_errors();
+    .unwrap();
 
     let mut device = AstarteDeviceSdk::new(&sdk_options).await.unwrap();
     let rx_data_ind_datastream = Arc::new(Mutex::new(HashMap::new()));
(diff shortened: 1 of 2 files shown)


tests/e2etest/main.rs
```diff
@@ -613,12 +620,15 @@ async fn http_post_to_intf(
...
         .send()
         .await
-        .map_err(|e| format!("HTTP POST failure: {e}"))?
-        .text()
-        .await
-        .map_err(|e| format!("Failure in parsing the HTTP POST result: {e}"))?;
-    if response != value_json {
-        println!("Server response: {response}");
+        .map_err(|e| format!("HTTP POST failure: {e}"))?;
+    if response.status() != 200 {
```
FrancescoVaiani, line 624: Can we replace 200 with the Reqwest status code enum?
```suggestion
    if response.status() != StatusCode::Ok{
```

tests/e2etest/main.rs
```diff
@@ -647,12 +657,15 @@ async fn http_delete_to_intf(
...
         .send()
         .await
-        .map_err(|e| format!("HTTP POST failure: {e}"))?
-        .text()
-        .await
-        .map_err(|e| format!("Failure in parsing the HTTP POST result: {e}"))?;
-    if !response.is_empty() {
-        println!("Server response: {response}");
+        .map_err(|e| format!("HTTP DELETE failure: {e}"))?;
+    if response.status() != 204 {
```
FrancescoVaiani, line 661: Can we replace 204 with the Reqwest status code enum?
```suggestion
    if response.status() != StatusCode::NoContent{
```

FrancescoVaiani (changes requested):
I could not remember what the status code 204 stands for. 
Maybe we can begin to slowly replace the status codes with the Reqwest StatusCode enum.

sorru94:
> I could not remember what the status code 204 stands for. Maybe we can begin to slowly replace the status codes with the Reqwest StatusCode enum.

I noticed that in the `registration.rs` and `pairing.rs` modules we use `http::StatusCode` instead of `reqwest::StatusCode`.
I think it would be better to use the `reqwest::StatusCode` everywhere as we always compare to a result from `reqwest`.

harlem88 (approved):

PR #168, opened 2023-04-11
ci(test): add ci workflow to test on windows

### LINKED ISSUE #37: Test Windows support
A CI running on Windows should be added to check (and fix) errors preventing the SDK to be run under Windows

joshuachp:
Add the ci workflow to test on windows-latest. Use vcpkg to install openssl and set VCPKG_ROOT to the vcpkg installation path. This takes the rust-openssl ci.yml as example.

Closes #37

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,10 +16,13 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-
 version: 2
 updates:
-- package-ecosystem: cargo
-  directory: "/"
-  schedule:
-    interval: weekly
+    - package-ecosystem: github-actions
+      directory: /
+      schedule:
+          interval: weekly
+    - package-ecosystem: cargo
+      directory: "/"
+      schedule:
+          interval: weekly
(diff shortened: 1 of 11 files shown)


joshuachp:
Added `tempfile` to have an easy cross-platform way to create a unique temporary directory. We could have used `std::env::temp_dir`, but `tempfile` provides an easier way to make the path unique across tests

harlem88:
1. Leave only the commit for windows.
2. Add the windows CI below the build-workflow.yaml CI.
3. Try to run the e2etest CI also on windows, so add it into e2e-test.yaml.

joshuachp:
> 1. Leave only the commit for windows.  
> 2. Add the windows CI below the build-workflow.yaml CI.
> 3. Try to run the e2etest CI also on windows, so add it into e2e-test.yaml.

Unfortunately, the e2e test requires helm/kind-action which only supports Linux at the moment, and I cannot think of any easy way around this.

I moved all the CI test in a separate file to differentiate them from the build/checks. 

I also added the permission and concurrency directive and update the existing actions to the latest version. It's all in this PR since I was already working on the CI workflows.

joshuachp:
@harlem88 rebase done

harlem88 (approved):

Issue #85, opened 2022-12-05
Improve Checks on Dynamic Introspection


harlem88:
At this time, if you add an interface that is already in introspection, it is replaced by the new one, without checking the major  or minor version.

We should improve checking in case of :
- `new major > old major` :  error should be raise (can't upgrade an interface major at runtime) ?
- `same major` : error should be raise (Interface already present in mapping) ? or simple warning ? 
- `same minor` : error should be raise (Interface already present in mapping) ? or simple warning ? 
- `new minor < old minor` : error should be raise (can't downgrade an interface at runtime)
- `new minor > old minor` : the new interface replaces the old one

sorru94:
As a suggestion while resolving this issue evaluate if the struct `Interface` should be present in the public API of the SDK.

PR #183, opened 2023-05-15
 feat(introspection): add validation for introspection version 

joshuachp:
Validate that an interface has a grater version than the previous one.

Closes https://github.com/astarte-platform/astarte-device-sdk-rust/issues/85

Follows https://github.com/astarte-platform/astarte-device-sdk-rust/pull/169


src/interface/validation.rs
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    /// Validate if an interface is given the previous version `prev`.
+    ///
+    /// It will check whether:
+    ///
+    /// - Both the versions are valid
+    /// - The name of the interface is the same
+    /// - The new version is a valid successor of the previous version.
+    pub fn validate_with(&self, prev: &Self) -> Result<&Self, ValidationError> {
```
sorru94, line 112: You could validate if the next interface version is valid against the current. It would result in a more logical type of check.

src/interface/validation.rs
```diff
@@ -0,0 +1,284 @@
...
+        // Check that both the interfaces are valid
+        self.validate().map_err(ValidationError::InvalidNew)?;
+        prev.validate().map_err(ValidationError::InvalidPrev)?;
```
sorru94, line 115: We could assume the previous version is already valid. How could it have been installed otherwise?

src/interface/validation.rs
```diff
@@ -0,0 +1,284 @@
...
+    /// Validate if an interface is given the previous version `prev`.
+    ///
+    /// It will check whether:
+    ///
+    /// - Both the versions are valid
+    /// - The name of the interface is the same
+    /// - The new version is a valid successor of the previous version.
+    pub fn validate_with(&self, prev: &Self) -> Result<&Self, ValidationError> {
+        // Check that both the interfaces are valid
+        self.validate().map_err(ValidationError::InvalidNew)?;
```
sorru94, line 114: The `validate` method is called each time an interface is created with `from_str` or `from_file`.
Validating it again here might be redundant.

src/interface/validation.rs
joshuachp: Yes absolutely, if its not a problem i would like to keep it until #186 PR because for now you can still generate an invalid interface with serde_json::from_str

src/interface/validation.rs
joshuachp: With this check my intention was to validate self and not prev. Because I care about self being valid and not previous. So the method is implemented on the next version.

When using it looks like line the following:
https://github.com/astarte-platform/astarte-device-sdk-rust/blob/0a3e69457761036602f6a6aeb9fd3e602271f78e/src/options.rs#L180-L186

src/interface/validation.rs
joshuachp: Fixed in https://github.com/astarte-platform/astarte-device-sdk-rust/pull/186

src/interface/validation.rs
```diff
@@ -0,0 +1,284 @@
...
+            self.next_major, self.next_minor, self.prev_major, self.prev_minor
+        )
+    }
+}
+
+impl Interface {
+    /// Validate if an interface is valid
+    pub fn validate(&self) -> Result<(), Error> {
+        // TODO: add additional validation
+        if self.get_version_major() == 0 && self.get_version_minor() == 0 {
```
noaccOS, line 99: This could also be written as

```suggestion
        let version = (self.get_version_major(), self.get_version_minor());
        if version == (0, 0) {
```

which in my opinion is a bit more readable

joshuachp: Done

src/options.rs
```diff
@@ -142,29 +148,45 @@ impl AstarteOptions {
...
-
-        Ok(self)
+        let interface_files = std::fs::read_dir(Path::new(interfaces_directory))?
+            .map(|res| res.map(|entry| entry.path()))
+            .collect::<Result<Vec<_>, io::Error>>()?;
+
+        interface_files
+            .iter()
+            .filter(|path| path.is_file() && path.extension() == Some(OsStr::new("json")))
+            .try_fold(self, |acc, path| acc.interface_file(path))
```
harlem88: Perhaps these two blocks could be combined into one.

src/options.rs
```diff
@@ -88,6 +88,20 @@ pub struct AstarteOptions {
     pub(crate) keepalive: std::time::Duration,
 }
 
+impl Debug for AstarteOptions {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("AstarteOptions")
+            .field("realm", &self.realm)
+            .field("device_id", &self.device_id)
+            .field("credentials_secret", &self.credentials_secret)
```
harlem88, line 97: Remove the `credentials_secret` field from the output or obfuscate it, we should avoid logging sensitive data.

harlem88 (changes requested):

src/options.rs
joshuachp: Done, thanks

sorru94 (approved):

harlem88 (approved):

PR #188, opened 2023-05-31
 test(mock): introduce mocking for the AstarteSDK

### LINKED ISSUE #180: Improve code coverage

joshuachp:
Mock the AsyncClient and EventLoop of the SDK, so we can unit test the
SDK without having to connect to a broker.

Closes #180

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -202,7 +202,7 @@ pub enum AstarteError {
 }
 
 /// Payload format for an Astarte device event data.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 pub enum Aggregation {
     /// Individual data, can be both from a datastream or property.
     Individual(AstarteType),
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -145,7 +150,10 @@ pub enum AstarteError {
     ConnectionError(#[from] rumqttc::ConnectionError),
 
     #[error("malformed input from Astarte backend")]
-    DeserializationError,
+    DeserializationError(#[from] bson::de::Error),
+
+    #[error("malformed input from Astarte backend, missinv value 'v' in document: {0}")]
```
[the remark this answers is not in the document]
joshuachp: Done

harlem88 (approved):

src/lib.rs
```diff
@@ -1427,4 +1459,187 @@ mod test {
             panic!("Deserialization in not individual");
         }
     }
+
+    #[tokio::test]
+    async fn wait_for_connack() {
```
sorru94, line 1464:
```suggestion
    async fn test_wait_for_connack() {
```

src/mock.rs
```diff
@@ -0,0 +1,41 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! Mocks for the Astarte Device SDK.
+
+use mockall::mock;
+use rumqttc::{ClientError, ConnectionError, Event, MqttOptions, QoS};
+
+mock!(
+    pub AsyncClient {
+        // NOTE: This is mocked for compatibility, but it will be hard to test
```
sorru94, line 26: What does this comment mean?
The call to `new ` is not in a difficult-to-test spot. Mocking `pairing::get_transport_config` would be sufficient to test the arguments and the return values of this function.
Even if this function was hard to test, this happens often, no need to specify it above each function.

joshuachp: We would need to serialize the tests or have a lock on the tests, since this is a static method. See the mock all [synchronization example](https://github.com/asomers/mockall/blob/master/mockall/examples/synchronization.rs).

I put the comment since the function is never actually used and would be hard to use in the tests (hard to test).

It is called in the AstarteSDK::new, and I didn't write the test for the above reason and because it's a pretty straight forward function.

src/lib.rs
joshuachp: Done

src/mock.rs
joshuachp: Comment removed

sorru94: True, the effort needed to write the test since it's a very short and straightforward function.
However, I would still remove the comment.

sorru94 (approved):

harlem88 (approved):

PR #190, opened 2023-06-01
 feat(msrv): update the MSRV 1.59.0 -> 1.66.1 

### LINKED ISSUE #189: Bump Rust version to 1.66.1

joshuachp:
Update the MSRV to 1.66.1 in the Cargo.toml and CI files, and update dependencies.

Closes #189


joshuachp:
The only dependency kept back is rustls, since the rumqttc client still doesn't support version 
0.21 of rustls.

```
rustls                    0.20.8  0.20.8     0.21.1  0.20.8  incompatible
```

We will need to update it once it's updated upstream.

PR #192, opened 2023-06-06
refactor(deps): update dependencies and clean up features

### LINKED ISSUE #191: Multiple dependencies for cryptography libraries and async runtimes
We are dependent on both rustls and openssl, tokio and actix.

The for the first one we are using openssl only to generate the certificates in `src/crypto.rs`. This can be done by rust crates so we we can drop openssl as a direct dependency.

The actix runtime was a feature of sqlx, but we are not actually using it and can be fixed by changing the `runtime-actix-rustls -> runtime-tokio-rustls`.

joshuachp:
Make all the dependencies use rustls for TLS, still using openssl to generate the keys. Remove unused features from tokio to slim down the crate.

Closes #191

### DIFF
--- a/src/crypto.rs
+++ b/src/crypto.rs
@@ -18,8 +18,18 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+//! Crypto module to generate the CSR to authenticate the device to the Astarte.
+
 use std::str::FromStr;
 
+#[cfg(feature = "openssl")]
+use openssl::{
+    ec::{EcGroup, EcKey},
+    hash::MessageDigest,
+    nid::Nid,
+    pkey::PKey,
+    x509::{X509NameBuilder, X509ReqBuilder},
+};
 use p384::{
     ecdsa::{DerSignature, SigningKey},
     pkcs8::{EncodePrivateKey, LineEnding},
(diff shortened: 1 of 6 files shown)


harlem88 (commented):
Do you run some benchmarks on new certificate generation?
- Measure the execution time of `openssl_key` function, between this version and previous one.

joshuachp:
For security reasons, the generation of the certificate should be constant time, so it shouldn't make any performance difference, while giving a Rust implementation. 

Still, I will make a benchmark in a different repo to test it out.

joshuachp:
Here is the benchmark #201

src/crypto.rs
```diff
@@ -18,24 +18,103 @@
...
+#[derive(Debug)]
+pub struct Bundle {
+    pub private_key: PrivateKey,
+    /// PEM encoded CSR
+    pub csr: String,
+}
 
 impl Bundle {
-    pub fn new(cn: &str) -> Result<Bundle, ErrorStack> {
+    pub fn new(realm: &str, device_id: &str) -> Result<Bundle, Error> {
```
sorru94, line 71: Why keep both implementations available at all times?
The implementation can be simpler if only one of them is active.
The two functions could be renamed to have the same name (`generate_key`).
The directive `#[cfg(not(feature = "openssl"))]` can be used to compile only the version that is selected.
If a user wants to use `openssl` it will most likely use it always, so no need to provide both methods.

src/lib.rs
```diff
@@ -19,7 +19,7 @@
  */
 #![doc = include_str!("../README.md")]
 
-mod crypto;
+pub mod crypto;
```
sorru94, line 22: Why make all of the content of crypto public? The user should need only the errors, right?

sorru94 (changes requested):

src/crypto.rs
```diff
@@ -18,24 +18,103 @@
...
+        if cfg!(feature = "openssl") {
+            #[cfg(feature = "openssl")]
+            return Self::openssl_key(realm, device_id);
+        }
```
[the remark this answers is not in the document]
joshuachp: Done

src/lib.rs
joshuachp: done

harlem88 (approved):

PR #202, opened 2023-06-26
fix(ci): update workflows to run on master and release for push

joshuachp:


joshuachp:
Following https://github.com/edgehog-device-manager/edgehog-device-runtime/pull/236

PR #213, opened 2023-07-03
feat(serialize/payload): improve ser/de and control payload

joshuachp:
Use Serialize and Deserialize to not allocate an additional HashMap when working with the payload. Separate the types errors int the specific module. Check the length of the control payload after deflating it and remove the unwraps.

### DIFF
--- a/src/payload.rs
+++ b/src/payload.rs
@@ -1,0 +1,272 @@
+// This file is part of Astarte.
+//
+// Copyright 2023 SECO Mind Srl
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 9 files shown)


src/payload.rs
```diff
@@ -0,0 +1,269 @@
...
+
+            Ok(Aggregation::Object(hmap))
+        }
+        value => {
+            let individual = AstarteType::try_from(value)?;
+
+            Ok(Aggregation::Individual(individual))
+        }
+    }
+}
```
harlem88, line 124: Move this function below `fn serialize_object`.

harlem88 (changes requested):

src/payload.rs
rgallor, line 1: Some pub functions/methods are missing doc comments.

src/properties.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    let (size, data) = bdata.split_at(4);
+    // The size is a u32 in big endian, so we need to convert it to usize
+    let size: u32 = u32::from_be_bytes([size[0], size[1], size[2], size[3]]);
+    let size: usize = size.try_into()?;
+
+    let mut d = ZlibDecoder::new(data);
+    let mut s = String::new();
+    let bytes_red = d.read_to_string(&mut s)?;
```
rgallor, line 56: should be `bytes_read` (not `red`).

rgallor (changes requested):

sorru94 (approved):

src/payload.rs
joshuachp: Done

src/properties.rs
joshuachp: Fixed

rgallor (approved):

PR #214, opened 2023-07-03
 refactor(interfaces): return and pass references when possible 

joshuachp:
The database trait now requires an AstarteType ref. The instructions in the event handler have been reordered to avoid cloning the payload when not needed. This also improves the deserialization which is done only once. Cleaned up some functions that are unused.

### DIFF
--- a/src/database.rs
+++ b/src/database.rs
@@ -61,7 +61,7 @@ pub trait AstarteDatabase {
         &self,
         interface: &str,
         path: &str,
-        value: &[u8],
+        value: &AstarteType,
         interface_major: i32,
     ) -> Result<(), Error>;
     /// Load a property from the database.
@@ -86,7 +86,7 @@ impl AstarteDatabase for AstarteSqliteDatabase {
         &self,
         interface: &str,
         path: &str,
-        value: &[u8],
+        value: &AstarteType,
         interface_major: i32,
     ) -> Result<(), Error> {
         debug!(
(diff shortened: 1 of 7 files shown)


src/error.rs
```diff
@@ -23,38 +23,23 @@ use std::convert::Infallible;
...
 /// Astarte error.
 ///
 /// Possible errors returned by functions of the Astarte device SDK.
 #[non_exhaustive]
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
-    #[error("bson serialize error")]
-    BsonSerError(#[from] bson::ser::Error),
-
     #[error("bson client error")]
```
rgallor, line 37: Missing doc (also other errors)

src/interface/error.rs
rgallor, line 1: Missing doc for error types. 
Also, to use the same pattern of what have been done in other errors, separate each error with a white space.

rgallor (changes requested):

src/interface/error.rs
joshuachp: Added

src/database.rs
```diff
@@ -223,7 +220,7 @@ mod test {
         //non existing
         assert_eq!(db.load_prop("com.test", "/test", 1).await.unwrap(), None);
 
-        db.store_prop("com.test", "/test", &ser, 1).await.unwrap();
```
sorru94, line 226: `ser` is not used anymore and could be removed.

src/interface/error.rs
```diff
@@ -31,10 +31,10 @@ pub enum InterfaceError {
     Io(#[from] io::Error),
     #[error("wrong major and minor")]
     MajorMinor,
-    #[error("interface not found")]
-    InterfaceNotFound,
-    #[error("mapping not found")]
-    MappingNotFound,
+    #[error("couldn't fine interface '{name}'")]
```
sorru94, line 34:
```suggestion
    #[error("couldn't find interface '{name}'")]
```

src/database.rs
joshuachp: Ser is still used at the end of the test

harlem88 (approved):

sorru94 (approved):

rgallor (approved):

PR #215, opened 2023-07-03
 feat(store): make the device SDK generic over the store 

joshuachp:
- Make the device SDK require a store, in memory or sqlite, and provide type aliases for the most common cases.
- Created the in memory store so that the device SDK can be used without a database, while the property would be kept until the device is restarted.
- Make the trait PropertyStore have a type parameter Err so that can be implemented outside the crate.
- Create a StorageWrapper that can be used to wrap a store and uniform the Err type.
- In the Sqlite store use the migrations and macros to bootstrap the database locally, or have the sqlx-data.json for compile time query validation.
- Fixed the primary keys of the storage to be non null, since in sqlite they are nullable by default.


src/options.rs
```diff
@@ -119,25 +119,45 @@ impl AstarteOptions {
...
-    /// Add a database to the astarte options.
-    pub fn database<T: AstarteDatabase + 'static + Sync + Send>(mut self, database: T) -> Self {
-        self.database = Some(Arc::new(database));
-        self
+impl<S> AstarteOptions<S>
+where
+    S: PropertyStore,
+{
+    /// Add a database to the Astarte options.
+    pub fn database<T>(self, database: T) -> AstarteOptions<T> {
```
rgallor, line 146: Shouldn't be more correct to use `store` rather then `database`?

src/error.rs
```diff
@@ -71,13 +69,18 @@ pub enum Error {
...
     /// Errors that can occur handling the payload.
     #[error("couldn't process payload")]
     Payload(#[from] PayloadError),
 
+    /// Error while parsing the /control/consumer/properties payload.
     #[error("couldn't handle properties")]
     Properties(#[from] PropertiesError),
+
+    /// Error returned by a store operation.
+    #[error("could't not complete store operation")]
```
harlem88, line 84:
```suggestion
    #[error("could't complete store operation")]
```

src/store/sqlite.rs
```diff
@@ -0,0 +1,245 @@
...
+use crate::{
+    payload::{self, Payload, PayloadError},
+    types::AstarteType,
+};
+
+/// Error returned by the [`SqliteStore`]
+#[non_exhaustive]
+#[derive(Debug, thiserror::Error)]
+pub enum SqliteError {
+    /// Error returned when the database uri is not valid
```
rgallor, line 37: Missing "." at the end of the doc comments

rgallor (changes requested):

harlem88 (changes requested):

src/error.rs
joshuachp: Done

src/store/sqlite.rs
joshuachp: Done

src/lib.rs
```diff
@@ -595,20 +597,20 @@ impl AstarteDeviceSdk {
         Ok(())
     }
 
-    /// When present get property from the allocated database (if allocated).
+    /// When present get property from the allocated storage (if allocated).
```
sorru94, line 600:
```suggestion
    /// Get property, when present, from the allocated storage.
```

src/options.rs
```diff
@@ -119,25 +119,45 @@ impl AstarteOptions {
...
 
-    /// Add a database to the astarte options.
-    pub fn database<T: AstarteDatabase + 'static + Sync + Send>(mut self, database: T) -> Self {
-        self.database = Some(Arc::new(database));
-        self
+impl<S> AstarteOptions<S>
+where
+    S: PropertyStore,
+{
+    /// Add a store to the Astarte options.
```
sorru94, line 145:
```suggestion
    /// Add a form of storage to the Astarte options.
```

src/store/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    /// Stores a property within the database.
+    async fn store_prop(
+        &self,
+        interface: &str,
+        path: &str,
+        value: &AstarteType,
+        interface_major: i32,
+    ) -> Result<(), Self::Err>;
+    /// Load a property from the database.
```
sorru94, line 56: I would maybe specify in this docs that it might also delete the property from the database if a version mismatch is detected.

src/store/sqlite.rs
```diff
@@ -0,0 +1,245 @@
...
+    async fn delete_prop(&self, interface: &str, path: &str) -> Result<(), Self::Err> {
+        sqlx::query_file!("queries/delete_prop.sql", interface, path)
+            .execute(&self.db_conn)
+            .await?;
+
+        Ok(())
+    }
+
+    async fn clear(&self) -> Result<(), Self::Err> {
+        sqlx::query!("delete from propcache")
```
sorru94, line 213: Shouldn't this use the relative query file?

src/store/sqlite.rs
```diff
@@ -0,0 +1,245 @@
...
+}
+
+/// Result of the load_prop query
+#[derive(Debug, Clone)]
+struct PropRecord {
+    value: Vec<u8>,
+    interface_major: i32,
+}
+
+/// Result of the load_prop query
```
sorru94, line 65: This should be from the `load_all_props` query, right?

sorru94 (changes requested):

src/lib.rs
```diff
@@ -442,31 +452,29 @@ impl AstarteDeviceSdk {
...
+                .expect("load_prop failed")
+                .expect("property wasn't correctly saved in the storage");
+
+            assert_eq!(*data, stored_prop);
+            let prop = self
+                .property(interface_name, path)
+                .await?
+                .expect("property wasn't correctly saved in the storage");
+            assert_eq!(*data, prop);
+            trace!("storage test ok");
```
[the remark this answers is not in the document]
harlem88: I think so too, it is a simple test to see if a property has been stored correctly, so it can be accomplished by a unit test.

src/store/sqlite.rs
```diff
@@ -0,0 +1,245 @@
...
+///
+/// The values are stored as a BSON serialized SQLite BLOB. That can be then deserialized in the
+/// respective [`AstarteType`].
+#[derive(Clone, Debug)]
+pub struct SqliteStore {
+    db_conn: sqlx::SqlitePool,
+}
+
+impl SqliteStore {
+    /// Creates an sqlite database for the Astarte device.
```
harlem88, line 101:
```suggestion
    /// Creates a sqlite database for the Astarte device.
```

harlem88 (changes requested):

CHANGELOG.md
```diff
@@ -8,13 +8,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Support for different case conventions on `AstarteAggregate` derive macro
   ([#126](https://github.com/astarte-platform/astarte-device-sdk-rust/issues/126)).
+- Add support for in-memory `MemoryStore` properties if no database is provided
+- Make `AstarteDeviceSdk` generic over the store type
```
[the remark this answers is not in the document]
joshuachp: Done

CHANGELOG.md
```diff
@@ -8,13 +8,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Support for different case conventions on `AstarteAggregate` derive macro
   ([#126](https://github.com/astarte-platform/astarte-device-sdk-rust/issues/126)).
+- Add support for in-memory `MemoryStore` properties if no database is provided
```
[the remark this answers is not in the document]
joshuachp: Done

src/options.rs
joshuachp: Changed, I made it clearer

src/lib.rs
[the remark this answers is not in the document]
joshuachp: This is a check at runtime that would yield little value in the unit test. Since the device is already tested, but could catch some bugs during development. Maybe if we separate it in a different function?

src/store/mod.rs
joshuachp: That is implementation specific, I will add a SHOULD

harlem88 (approved):

sorru94 (approved):

harlem88 (approved):

PR #216, opened 2023-07-03
 fix(type): panic when deserializing empty bson array 

joshuachp:
When deserializing an empty bson array, we whould access the first element to know the type. This is not possible when the array is empty, and it would panic.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -86,9 +86,9 @@ pub enum AstarteType {
     BinaryBlobArray(Vec<Vec<u8>>),
     DateTimeArray(Vec<DateTime<Utc>>),
 
-    /// A generic empty array. This is out of the mqtt-v1-protocol, but it's needed since we do not
-    /// know the type of the array without elements. It should only be build when converting from a
-    /// bson array.
+    /// A generic empty array. This is not part of the Astarte MQTT v1 protocol. However, it's
+    /// required since we cannot know the type of an empty array. It should only be used when
+    /// converting from a Bson array.
     EmptyArray,
 
     Unset,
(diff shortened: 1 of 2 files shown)


rgallor (approved):

src/types.rs
```diff
@@ -300,108 +321,101 @@ impl From<AstarteType> for Bson {
...
             Bson::Null => Ok(AstarteType::Unset),
+            Bson::Array(arr) => AstarteType::from_bson_array(arr),
             _ => Err(Self::Error::FromBsonError(format!(
                 "Can't convert {d:?} to astarte"
             ))),
         }
     }
 }
 
+/// Utility to convert a Bson array into a AstarteType array
```
rgallor, line 364: ... into "an" AstarteType array.

rgallor (changes requested):

src/types.rs
joshuachp: Done

harlem88 (approved):

harlem88 (changes requested):
Add a line about the fix on the Changelog

rgallor (approved):

src/types.rs
```diff
@@ -86,31 +86,52 @@ pub enum AstarteType {
     BinaryBlobArray(Vec<Vec<u8>>),
     DateTimeArray(Vec<DateTime<Utc>>),
 
+    /// A generic empty array. This is out of the mqtt-v1-protocol, but it's needed since we do not
+    /// know the type of the array without elements. It should only be build when converting from a
+    /// bson array.
```
sorru94, line 91:
```suggestion
    /// A generic empty array. This is not part of the Astarte MQTT v1 protocol. However, it's
    /// required since we cannot know the type of an empty array. It should only be used when
    /// converting from a Bson array.
```

sorru94 (approved):

harlem88 (approved):

PR #226, opened 2023-07-10
fix(unset): send an empty buffer on property unset

joshuachp:
This is the documented way on how to unset a property. Before we were sending a bson document with a null value which was also working, but it's not the way described in the mqtt-v1-protocol.


rgallor (approved):

harlem88:
Add an entry to the changelog about this change.

sorru94 (approved):

harlem88 (approved):

PR #229, opened 2023-07-27
Feat event loop

### LINKED ISSUE #227: Feat!: event_loop with channel and reconnect with exponential backoff
In the `rumqttc` client, the reception, dispatch of messages and reconnection is done through the `eventloop.poll().await` method.

At the current time, we are not always advancing the event loop; since we are returning the incoming messages to the users of the library. This could lead to problems while multiple threads are sending messages while we are handling an incoming payload, since they wouldn't be sent until the payload is handled. Furthermore, the MQTT library implements a reconnection strategy, but gives the user the liberty on how to handle it. So this means that at the moment we are not implementing an exponential back off for the loop, and we would spam errors whenever the client is disconnected from the broker.

#### Proposal

The idea I would like to propose is a breaking change from the current API (but we could still manage to make it compatible). The main idea is that the `AstarteSdk::handle_events` function will keep looping, with the only responsibility to handle the connection (reconnect and sending packages). The incoming messages would then be moved each to a separate task and be then accessible from outside the library through a channel present in the AstarteSDK struct.

For the reconnection in the event loop we can match on the error and use `tokio::time` each iteration to wait for the reconnection.

joshuachp:
Closes #227

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Handle MQTT connection errors by trying to reconnect.
+- Make the reconnection attempt wait with an exponential back-off.
+
+### Changed
+- Return a channel for the events when creating a device SDK.
+- Make handle event loop block to handle the events.
+- Create a shareable struct (`Arc`) of the `AstarteDeviceSdk` to not clone the
+  device id and realm.
+- Make the `DynError` trait bound shareable across threads.
+
 ### Fixed
 - Unset of property send empty buffer instead of document with null value.
 
--- a/examples/individual_datastream/main.rs
+++ b/examples/individual_datastream/main.rs
@@ -52,11 +52,12 @@ async fn main() -> Result<(), Error> {
     .ignore_ssl_errors();
 
     // Create an Astarte Device (also performs the connection)
-    let mut device = astarte_device_sdk::AstarteDeviceSdk::new(sdk_options).await?;
+    let (mut device, mut rx_events) =
+        astarte_device_sdk::AstarteDeviceSdk::new(sdk_options).await?;
+    let device_cpy = device.clone();
     println!("Connection to Astarte established.");
 
-    // Create an thread to transmit
-    let device_cpy = device.clone();
+    // Create an task to transmit
     tokio::task::spawn(async move {
         loop {
             // Send endpoint 1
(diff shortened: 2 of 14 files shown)


examples/individual_datastream/main.rs
```diff
@@ -52,11 +52,12 @@ async fn main() -> Result<(), Error> {
...
     // Create an Astarte Device (also performs the connection)
-    let mut device = astarte_device_sdk::AstarteDeviceSdk::new(sdk_options).await?;
+    let (mut device, mut rx_events) =
+        astarte_device_sdk::AstarteDeviceSdk::new(sdk_options).await?;
+    let device_cpy = device.clone();
     println!("Connection to Astarte established.");
 
-    // Create an thread to transmit
-    let device_cpy = device.clone();
+    // Create an task to transmit
```
harlem88, line 60: `an` -> `a`

src/error.rs
```diff
@@ -41,6 +41,10 @@ pub enum Error {
     #[error("mqtt connection error")]
     ConnectionError(#[from] rumqttc::ConnectionError),
 
+    /// The e connection poll reached the max number of retries.
```
harlem88, line 44: Remove `e` -> `The connection poll`...

harlem88 (changes requested):

examples/individual_datastream/main.rs
joshuachp: Done

src/error.rs
joshuachp: Done

src/retry.rs
```diff
@@ -0,0 +1,96 @@
...
+/// Iterator that yields a delay that will increase exponentially till the max,
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct DelaiedPoll {
+    max: u32,
+    delay: u32,
+    exp: u32,
+    base: u32,
+}
+
+impl DelaiedPoll {
```
sorru94, line 37: `DelayedPoll`

sorru94 (changes requested):

src/retry.rs
joshuachp: Done

joshuachp: Done

sorru94 (approved):

src/lib.rs
```diff
@@ -188,25 +202,32 @@ where
...
+    pub async fn new(
+        opts: AstarteOptions<S>,
+    ) -> Result<(AstarteDeviceSdk<S>, EventReceiver), Error> {
         let mqtt_options = pairing::get_transport_config(&opts).await?;
 
         debug!("{:#?}", mqtt_options);
 
         let (client, eventloop) = AsyncClient::new(mqtt_options, 50);
 
+        let (tx_events, rx_events) = mpsc::channel(50);
```
harlem88, line 214: If `50` refers to the same think (specifies the capacity of the bounded channel) maybe should be defined as constant.
```suggestion
        let (client, eventloop) = AsyncClient::new(mqtt_options, MQTT_CHANNEL_SIZE);

        let (tx_events, rx_events) = mpsc::channel(MQTT_CHANNEL_SIZE);
```

harlem88 (changes requested):
Since the return of `AstarteDeviceSdk::new()` function has changed, describe it as a Changelog entry.

src/lib.rs
joshuachp: Done

CHANGELOG.md
```diff
@@ -5,6 +5,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+### Added
+- Handle MQTT connection errors by trying to reconnect
+- Make the reconnection attempt wait with an exponential back-off
+
+### Changed
+- Return a channel for the events when creating a device SDK
+- Make handle event loop block to handle the events
+- Create a shareable struct (`Arc`) of the `AstarteDeviceSdk` to not clone the
+  device id and realm
+- Make the `DynError` trait bound shareable across threads
```
harlem88, line 17:
```suggestion
- Handle MQTT connection errors by trying to reconnect.
- Make the reconnection attempt wait with an exponential back-off.

### Changed
- Return a channel for the events when creating a device SDK.
- Make handle event loop block to handle the events.
- Create a shareable struct (`Arc`) of the `AstarteDeviceSdk` to not clone the
  device id and realm.
- Make the `DynError` trait bound shareable across threads.
```

harlem88 (changes requested):

PR #236, opened 2023-09-11
feat(derive): update the aggregate derive macro

joshuachp:
- Update `syn` to version 2.0 and refactor the code to support the breaking changes for the new major version.
- Restructure the code to use the `Parse` trait to get the information from the `TokenStream`.
- Do not panic, but return a compile error through the `syn::Error` which are converted in `compile_error!(..)`.
- Check the name and value from the derive attribute, before only the value was checked.
- Require the generics on the struct to implement `TryInto<AstarteType>` so so we can have generics in the aggregate.
- Use `quote_spanned` for the conversion for better error messages.
- Use the full path `astarte_device_sdk::AggregateDerive` for the trait impl to prevent conflicts with the trait.


joshuachp:
> The changes regarding the derive crate look good. But I don't understand why the workspace commit (`1eb105ff9d1e662343b3674d72b0e42136546728`) is present in this PR.

FTR: the workspace is needed to make the CI run also for the derive crate, before was ignored

sorru94 (approved):

.cargo/config.toml
```diff
@@ -0,0 +1,20 @@
+# This file is part of Edgehog.
```
harlem88, line 1:
```suggestion
# This file is part of Astarte.
```

harlem88 (changes requested):

.cargo/config.toml
joshuachp: Fixed

harlem88 (approved):

harlem88 (approved):

PR #240, opened 2023-09-26
fix(store): add correct Send + Sync + 'static for tokio

joshuachp:
This will let the error be passed to a tokio task. The error is overzealous on the Sync bound, but most of the methods of the Error trait requires it anyway so I think it's safe to add.

### DIFF
--- a/src/store/error.rs
+++ b/src/store/error.rs
@@ -19,9 +19,9 @@
 //! Error for the store.
 
 /// Dynamic error type of an [`super::PropertyStore`].
-type DynError = Box<dyn std::error::Error>;
+type DynError = Box<dyn std::error::Error + Send + Sync + 'static>;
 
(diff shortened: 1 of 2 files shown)


joshuachp:
This is a fix where the error cannot be sent across `tokio::spawn` when used outside the lib. This requires a new release of 0.6.1, and it would be better to yank the Astarte SDK version 0.6.0.

harlem88 (commented):
Fix [Code coverage tarpauline](https://github.com/astarte-platform/astarte-device-sdk-rust/actions/runs/6323322579/job/17170666348?pr=240#logs) in this PR or in a new one.
`cargo tarpaulin --engine llvm --ignore-tests --lib --out xml -- --test-threads 1`

joshuachp:
Fixed in #244

harlem88 (approved):

PR #242, opened 2023-09-27
fix(serde): serialize and deserialize mixed arrays

### LINKED ISSUE #238: Bug in reception of `longintegerarray`
When Astarte streams a `longintegerarray` to a device containing small values, the `handle_events` function returns `Payload(AstarteType(FromBsonArrayError))`.
To reproduce use the array: `[45543543534, 10, 0, 45543543534]`.

The issue seems to be originating in the [`from_bson_vec`](https://github.com/astarte-platform/astarte-device-sdk-rust/blob/4e426a3ed6d19de6ff657b228cc2ab6acc2f7048/src/types.rs#L377C12-L377C25) function.
The function operates under the assumption that Astarte will send a BSON array containing all `i64` when transmitting a `longintegerarray`. Instead, Astarte sends `i64` only for values not fitting in an `i32`, otherwise, it will send an `i32`.

Two issues arise:
1. The function assumes the size of each element of the array from the size of the first element. This would result in failure when the first element has a small value (let's say `10`) and one or more of the following elements has a much larger value (i.e. `45543543534`).
2. When dealing with a `longintegerarray` such as `[45543543534, 10, 0, 45543543534]` the function uses the closure `|b| b.as_i64()` over each element. The method `as_i64` of a `Bson` object expects only `i64` as inputs and returns none in all other cases, even when receiving an `i32`.

joshuachp:
Fixes a bug when deserializing integer arrays of mixed sizes received from astarte. Also improve the validation when sending and receiving data, introduced more ergonomic ways to access the interfaces mappings/objects. Upgraded to hard errors when interfaces are non-existing or of a different type.

Closes #238

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Check if an interface exists and the type is the same of the value
+  passed/received when sending or receiving data from Astarte.
+
 ### Fixed
 - Unset of property send empty buffer instead of document with null value.
+- Deserialize mixed integer BSON arrays from Astarte to the type specified in
+  the interface (longinteger and integer)
 
 ## [0.6.0] - 2023-07-05
 ### Added
--- a/src/interface/mapping/vec.rs
+++ b/src/interface/mapping/vec.rs
@@ -1,0 +1,218 @@
+// This file is part of Astarte.
+//
+// Copyright 2023 SECO Mind Srl
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+//! Sorted collection of interfaces accessible by endpoint or path.
+
+use std::{
+    cmp::Ordering, collections::BTreeSet, fmt::Debug, iter::FusedIterator, ops::Deref,
+    slice::Iter as SliceIter,
+};
+
+use itertools::Itertools;
+
+use super::{endpoint::Endpoint, InterfaceMapping};
+
+#[derive(Debug, Clone, Default)]
+pub(crate) struct MappingVec<T> {
+    mappings: Vec<Item<T>>,
+}
+
+impl<T> MappingVec<T> {
+    /// Create an empty vector.
+    pub(crate) fn new() -> Self {
+        Self {
+            mappings: Vec::new(),
+        }
+    }
+
+    /// Get's the mapping searching the [`Endpoint`]'s for the one matching the path.
+    pub(crate) fn get<U>(&self, path: &U) -> Option<&T>
+    where
+        T: InterfaceMapping + Debug,
+        U: PartialOrd<Endpoint<String>> + Eq + Debug,
+    {
+        self.mappings
+            .binary_search_by(|item| {
+                let endpoint = item.endpoint();
+                // We invert the ordering for less/greater since the trait is implemented on the
+                // path, but we need to actually check the argument
+                path.partial_cmp(endpoint)
+                    .map_or(Ordering::Less, Ordering::reverse)
+            })
+            .ok()
+            .and_then(|idx| self.mappings.get(idx))
+            .map(|item| &item.0)
+    }
+
+    /// Returns the number of mappings.
+    pub(crate) fn len(&self) -> usize {
+        self.mappings.len()
+    }
+
+    /// Iterate over the mappings.
+    pub(crate) fn iter(&self) -> ItemIter<T> {
+        ItemIter {
+            items: self.mappings.iter(),
+        }
+    }
+}
+
+impl<T> PartialEq for MappingVec<T>
+where
+    T: PartialEq,
+{
+    fn eq(&self, other: &Self) -> bool {
+        self.mappings
+            .iter()
+            .zip_longest(other.mappings.iter())
+            .all(|e| match e {
+                itertools::EitherOrBoth::Both(a, b) => a.0 == b.0,
+                itertools::EitherOrBoth::Left(_) | itertools::EitherOrBoth::Right(_) => false,
+            })
+    }
+}
+
+impl<T> Eq for MappingVec<T> where T: Eq {}
+
+impl<T> From<BTreeSet<Item<T>>> for MappingVec<T> {
+    fn from(value: BTreeSet<Item<T>>) -> Self {
+        let mappings = value.into_iter().collect();
+
+        Self { mappings }
+    }
+}
+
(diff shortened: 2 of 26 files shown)


src/interface/mod.rs
```diff
@@ -170,10 +172,31 @@ impl Interface {
...
     pub fn is_property(&self) -> bool {
         matches!(self.inner, InterfaceType::Properties(_))
     }
 
+    /// Returns a [`PropertyRef`] if the interface is a property.
+    pub(crate) fn as_prop_ref(&self) -> Option<PropertyRef> {
+        self.is_property().then_some(PropertyRef(self))
+    }
+
+    /// Check if the interface is a property.
```
harlem88, line 190: Check if the interface is an object.

src/validate.rs
```diff
@@ -0,0 +1,160 @@
...
+
+/// Errors returning while validating a send or a receive
+#[non_exhaustive]
+#[derive(thiserror::Error, Debug)]
+pub enum ValidationError {
+    /// Sending timestamp to a mapping without `explicit_timestamp`
+    #[error("sending timestamp to a mapping without `explicit_timestamp`")]
+    Timestamp,
+    /// Un-setting property without `allow_unset`
+    #[error("uset is only for property with `allow_unset`")]
```
harlem88, line 37: `uset` -> `unset`

harlem88 (changes requested):

src/interface/mod.rs
joshuachp: Corrected thanks

src/validate.rs
joshuachp: Corrected

src/interface/mod.rs
```diff
@@ -278,64 +304,54 @@ pub(crate) enum InterfaceType {
...
 impl DatastreamIndividual {
-    pub fn iter_mappings(&self) -> IndividualMappingIter {
+    pub(crate) fn iter_mappings(&self) -> IndividualMappingIter {
         IndividualMappingIter::new(&self.mappings)
     }
 
-    pub fn add_mapping(&mut self, mapping: &Mapping) -> Result<(), InterfaceError> {
-        let individual = DatastreamIndividualMapping::try_from(mapping)?;
+    pub(crate) fn add_mapping(
+        tree: &mut MappingSet<DatastreamIndividualMapping>,
```
harlem88, line 316: Why didn't use `&self` on this method? So you can put `tree` as internal method, without pass it as external argument.

harlem88 (changes requested):

src/interface/mod.rs
joshuachp: The tree is not internal to the struct, there are two types of structures MappingSet and MappingVec (the names are unfortunately similar), the tree is needed during contruction of the interface to ensure that the mappings are unique, while the vec is used to retrieve a mapping comparing the endpoints. 

This is a static method on Self since is dependent on the type of interface and is used to add a mapping to the three, when all mapping are added the MappingSet will then be converted into a MappingVec

src/validate.rs
```diff
@@ -0,0 +1,160 @@
...
+        let interface = Interface::from_str(DEVICE_DATASTREAM).unwrap();
+
+        let mapping = MappingRef::new(&interface, mapping!("/boolean_endpoint")).unwrap();
+
+        let res = validate_send_individual(mapping, &AstarteType::Boolean(true), &None);
+        assert!(res.is_ok());
+
+        let res = validate_send_individual(mapping, &AstarteType::Boolean(true), &Some(Utc::now()));
+        assert!(res.is_ok());
+    }
```
harlem88, line 159: Add validation tests for sending set/unset properties and for receiving values.

harlem88 (changes requested):

sorru94 (approved):

sorru94:
As a suggestion, the next time try to split large PRs in multiple commits, dividing them for functionality if possible. 
For example, most changes in the `src/interface/mapping/endpoint.rs` file seem to have a different aim compared to the main purpose of this PR.
I think it's fine to keep them in this PR, but maybe it would have been better to make incremental commits with smaller changes.

src/validate.rs
joshuachp: I will add the test for the unset in the deserialization function, since the check has been moved there.

joshuachp: Done

harlem88 (approved):

src/interface/mapping/vec.rs
```diff
@@ -0,0 +1,223 @@
...
+
+use super::{endpoint::Endpoint, InterfaceMapping};
+
+#[derive(Debug, Clone, Default)]
+pub(crate) struct MappingVec<T> {
+    mappings: Vec<Item<T>>,
+}
+
+impl<T> MappingVec<T> {
+    /// Create and empty vector.
```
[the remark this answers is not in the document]
joshuachp: Done

src/interface/mapping/vec.rs
```diff
@@ -0,0 +1,223 @@
...
+
+impl<T> From<BTreeSet<Item<T>>> for MappingVec<T> {
+    fn from(value: BTreeSet<Item<T>>) -> Self {
+        let mappings = value.into_iter().collect();
+
+        Self { mappings }
+    }
+}
+
+/// Item of the to check that the inner type is [`Ord`] by the [`Endpoint`] returned by
```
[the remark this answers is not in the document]
joshuachp: Added `[`MappingVec`]`

src/lib.rs
```diff
@@ -423,45 +421,75 @@ where
...
+    ) -> Result<Aggregation, Error> {
+        let interfaces = self.interfaces.read().await;
+        let interface = interfaces.get(interface).ok_or_else(|| {
+            warn!("publish on missing interface {interface} ({path})");
+            Error::MissingInterface(interface.to_string())
+        })?;
 
-                Ok(())
+        let (data, timestamp) = match interface.aggregation() {
+            // The interface is individual
```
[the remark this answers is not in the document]
joshuachp: Done

lucaato (approved):

harlem88 (approved):

src/interface/mapping/vec.rs
```diff
@@ -0,0 +1,223 @@
...
+    where
+        T: InterfaceMapping + Debug,
+        U: PartialOrd<Endpoint<String>> + Eq + Debug,
+    {
+        self.mappings
+            .binary_search_by(|item| {
+                let endpoint = item.endpoint();
+                // We invert the ordering for less/greater since the trait is implemented on the
+                // path, but we need to actually check the argument
+                if path.eq(endpoint) {
```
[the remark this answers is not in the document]
joshuachp: Even if this is practically the same, it's not fully correct. Since you're trying to order a PartialOrd sequence of items (eg. floats which have Nan) and giving them all the out of domain values (+Nan and -Nan that are out of R) the Ordering::Less. I would like to keep the comparison explicit like this, even if in practice it will never happen. Here is an example https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=501d604dd6db6c4ccd0414f79860fc7c

CHANGELOG.md
```diff
@@ -5,8 +5,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Check the presence and type of an interface when sending or receiving data
+  from Astarte.
+
 ### Fixed
 - Unset of property send empty buffer instead of document with null value.
+- Deserialize mixed integer BSON arrays from Astarte to the type type
```
harlem88, line 14:
```suggestion
- Deserialize mixed integer BSON arrays from Astarte to the type
```

harlem88 (changes requested):

CHANGELOG.md
joshuachp: Fixed

harlem88 (approved):

PR #251, opened 2023-10-10
fix(interface): make escaped strings deserializable

### LINKED ISSUE #250: Interfac Def cannot be deserialized with escaped characters
Since the `InterfaceDef` descriptions have type `&str` it cannot deserialize an escaped string which need to be unescaped.

It will result in the following error:
```
Parse(Error("invalid type: string \"Interface description \\\"escaped\\\"\", expected a borrowed string", line: 7, column: 62))
```

joshuachp:
Make the interface doc and description deserializable when there are escaped characters in the string.

Closes #250


harlem88 (commented):
Add an entry to the changelog about the this fix.

joshuachp:
Added

harlem88 (approved):

PR #260, opened 2023-12-12
Grpc integration

lucaato:
Implemented grpc transport and integrated message hub type conversions.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,7 @@ harness = false
 
 [dependencies]
 astarte-device-sdk-derive = { workspace = true, optional = true }
+astarte-message-hub-proto = { workspace = true, optional = true } 
 async-trait = { workspace = true }
 base64 = { workspace = true }
 bytes = { workspace = true }
@@ -79,14 +80,17 @@ env_logger = { workspace = true }
 mockall = { workspace = true }
 tempfile = { workspace = true }
 tokio = { workspace = true, features = ["rt", "rt-multi-thread"] }
+tokio-stream = { workspace = true, features = ["net"] }
 
 [features]
 derive = ["dep:astarte-device-sdk-derive"]
 openssl = ["dep:openssl"]
+message-hub = ["dep:astarte-message-hub-proto"]
 
 [workspace.dependencies]
 astarte-device-sdk-derive = { version = "0.6.1", path = "./astarte-device-sdk-derive" }
-async-trait = "0.1.41"
+astarte-message-hub-proto = "0.6.1"
+async-trait = "0.1.43"
 base64 = "0.21.0"
 bson = "2.1.0"
 bytes = "1.5.0"
@@ -96,11 +100,11 @@ ecdsa = "0.16.7"
 env_logger = "0.10.0"
 flate2 = "1.0.0"
 futures = "0.3.0"
-http = "0.2.0"
+http = "0.2.9"
 itertools = "0.11.0"
 log = "0.4.17"
 mockall = "0.11.4"
-once_cell = "1.13.1"
+once_cell = "1.17.1"
 openssl = "0.10.46"
 p384 = "0.13.0"
 proc-macro2 = "1.0.63"
@@ -119,6 +123,7 @@ syn = "2.0.31"
 tempfile = "3.6.0"
 thiserror = "1.0.35"
 tokio = "1.18.0"
+tokio-stream = "0.1.0"
 url = "2.2.2"
 uuid = "1.0.0"
 webpki = "0.22.0"
--- a/src/error.rs
+++ b/src/error.rs
@@ -115,4 +115,8 @@ pub enum Error {
 
     #[error(transparent)]
     MqttConnection(#[from] MqttConnectionError),
+
+    #[cfg(feature = "message-hub")]
+    #[error(transparent)]
+    GrpcTransport(#[from] crate::transport::grpc::GrpcTransportError),
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -52,6 +52,7 @@ pub use rumqttc;
 use async_trait::async_trait;
 use log::{debug, error, info, trace, warn};
 use tokio::sync::{mpsc, RwLock};
+use transport::Disconnect;
 
 /// Re-exported internal structs
 pub use crate::event::FromEvent;
@@ -68,7 +69,7 @@ use crate::store::wrapper::StoreWrapper;
 use crate::store::{PropertyStore, StoredProp};
 use crate::transport::{Publish, Receive, ReceivedEvent, Register};
 use crate::types::{AstarteType, TypeError};
-use crate::validate::{validate_individual, validate_object};
+use crate::validate::{ValidatedIndividual, ValidatedObject};
 
 /// A **trait** required by all data to be sent using
 /// [send_object()][crate::AstarteDeviceSdk::send_object] and
@@ -185,19 +186,15 @@ impl<S, C> AstarteDeviceSdk<S, C> {
 
     async fn handle_event(
         &self,
-        connection_event: &ReceivedEvent<C::Payload>,
+        interface: &str,
+        path: &str,
+        payload: C::Payload,
     ) -> Result<Aggregation, crate::Error>
     where
         S: PropertyStore,
         C: Receive + Sync,
     {
-        let ReceivedEvent {
-            interface,
-            path,
-            payload,
-        } = connection_event;
-
-        let path = MappingPath::try_from(path.as_str())?;
+        let path = MappingPath::try_from(path)?;
 
         let interfaces = self.interfaces.read().await;
         let interface = interfaces.get(interface).ok_or_else(|| {
@@ -226,7 +223,7 @@ impl<S, C> AstarteDeviceSdk<S, C> {
         &self,
         interface: &Interface,
         path: &MappingPath<'a>,
-        payload: &C::Payload,
+        payload: C::Payload,
     ) -> Result<(Aggregation, Option<chrono::DateTime<chrono::Utc>>), Error>
     where
         S: PropertyStore,
@@ -267,7 +264,7 @@ impl<S, C> AstarteDeviceSdk<S, C> {
         &self,
         interface: &Interface,
         path: &MappingPath<'a>,
-        payload: &C::Payload,
+        payload: C::Payload,
     ) -> Result<(Aggregation, Option<chrono::DateTime<chrono::Utc>>), Error>
     where
         S: PropertyStore,
@@ -382,9 +379,9 @@ impl<S, C> AstarteDeviceSdk<S, C> {
             return Ok(());
         }
 
-        let validated = validate_individual(mapping, path, &individual, timestamp)?;
+        let validated = ValidatedIndividual::validate(mapping, path, individual, timestamp)?;
 
-        self.connection.send_individual(validated).await?;
+        self.connection.send_individual(validated.clone()).await?;
 
         // Store the property in the database after it has been successfully sent
         // We just need to handle the Err case since the Ok was handled by returning early
@@ -397,7 +394,7 @@ impl<S, C> AstarteDeviceSdk<S, C> {
             let prop = StoredProp {
                 interface: interface_name,
                 path,
-                value: &individual,
+                value: &validated.into_data(),
                 interface_major,
                 ownership,
             };
@@ -438,7 +435,7 @@ impl<S, C> AstarteDeviceSdk<S, C> {
 
         let aggregate = data.astarte_aggregate()?;
 
-        let validated = validate_object(object, path, &aggregate, timestamp)?;
+        let validated = ValidatedObject::validate(object, path, aggregate, timestamp)?;
 
         self.connection.send_object(validated).await
     }
@@ -666,6 +663,11 @@ pub trait Client {
 }
 
 #[async_trait]
+trait ClientDisconnect {
+    async fn disconnect(self);
+}
+
+#[async_trait]
 impl<S, C> Client for AstarteDeviceSdk<S, C>
 where
     S: PropertyStore,
@@ -744,20 +746,28 @@ where
 
     async fn handle_events(&mut self) -> Result<(), crate::Error> {
         loop {
-            let event_payload = self.connection.next_event(&self.shared).await?;
+            let ReceivedEvent {
+                interface,
+                path,
+                payload,
+            } = self.connection.next_event(&self.shared).await?;
+
             let device = self.clone();
 
             tokio::spawn(async move {
-                let data = device
-                    .handle_event(&event_payload)
-                    .await
-                    .map(|aggregation| AstarteDeviceDataEvent {
-                        interface: event_payload.interface,
-                        path: event_payload.path,
-                        data: aggregation,
-                    });
+                let data =
+                    device
+                        .handle_event(&interface, &path, payload)
+                        .await
+                        .map(|aggregation| AstarteDeviceDataEvent {
+                            interface,
+                            path,
+                            data: aggregation,
+                        });
 
-                device.tx.send(data).await.expect("Channel dropped")
+                if let Err(e) = device.tx.send(data).await {
+                    error!("Error received while sending data: {}", e);
+                }
             });
         }
     }
@@ -807,6 +817,19 @@ where
         }
 
         self.remove_properties_from_store(interface_name).await
+    }
+}
+
+#[async_trait]
+impl<S, C> ClientDisconnect for AstarteDeviceSdk<S, C>
+where
+    S: PropertyStore,
+    C: Disconnect + Send,
+{
+    async fn disconnect(self) {
+        if let Err(e) = self.connection.disconnect().await {
+            warn!("Could not close the connection gracefully: {}", e);
+        }
     }
 }
 
--- a/src/transport/grpc/convert.rs
+++ b/src/transport/grpc/convert.rs
@@ -1,0 +1,1354 @@
+/*
+ * This file is part of Astarte.
+ *
+ * Copyright 2023 SECO Mind Srl
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Contains conversion traits to convert the Astarte types in the protobuf format to the
+//! Astarte types from the Astarte device SDK.
+
+use std::collections::HashMap;
+use std::num::TryFromIntError;
+use std::str::{FromStr, Utf8Error};
+
+use astarte_message_hub_proto;
+use astarte_message_hub_proto::{astarte_message::Payload as ProtoPayload, pbjson_types};
+use chrono::DateTime;
+
+use crate::{
+    transport::grpc::GrpcReceivePayload, transport::ReceivedEvent, types::AstarteType,
+    validate::ValidatedIndividual, validate::ValidatedObject, Aggregation, AstarteDeviceDataEvent,
+    Timestamp,
+};
+
+use super::GrpcTransportError;
+
+/// Error returned by the Message Hub types conversions.
+#[non_exhaustive]
+#[derive(thiserror::Error, Debug)]
+pub enum MessageHubProtoError {
+    /// Wrapper for integer conversion errors
+    #[error(transparent)]
+    TryFromIntError(#[from] TryFromIntError),
+
+    /// Expected field was not found
+    #[error("Missing the expected field '{0}'")]
+    ExpectedField(&'static str),
+
+    /// Cannot convert [`AstarteType::Unset`] to a proto [`IndividualData`]
+    #[error("Cannot perform conversion of an Unset to a proto IndividualData")]
+    UnsetConversion,
+
+    /// Conversion error while trying to convert the byte array into a string
+    #[error("The received byte array is not a valid utf8 string")]
+    ByteToUtf8StringConversion(#[from] Utf8Error),
+
+    /// Date conversion error
+    #[error("Error while converting a proto date: {0}")]
+    DateConversion(String),
+}
+
+/// Unwraps a variable chain of nested optionals by calling the [`Option::and_then`] assoctiated function
+/// The optional items can be properties or function that do not take parameters.
+/// Terminates the chain with an [`Option::ok_or`] method that returns a result
+/// # Examle:
+/// `optional_chain!(payload.take_data().take_individual().individual_data);`
+///
+/// This would compile to code similar to:
+/// ```rust
+/// use astarte_device_sdk::transport::grpc::convert::MessageHubProtoError;
+/// use astarte_message_hub_proto::astarte_message::Payload;
+/// use astarte_message_hub_proto::AstarteUnset;
+/// let payload = Payload::AstarteUnset(AstarteUnset {});
+/// payload.take_data()
+///     .and_then(|d| d.take_individual())
+///     .and_then(|d| d.individual_data)
+///     .ok_or(MessageHubProtoError::ExpectedField("payload.take_data().take_individual().individual_data"));
+/// ```
+macro_rules! optional_chain {
+    // callable from outside first token is a function ident
+    ($wrapper:ident.$func:ident()$($token:tt)*) => {{
+        optional_chain!(@subprop ($wrapper.$func()) ($wrapper.$func()) $($token)*)
+    }};
+
+    // callable from outside first token is an ident
+    ($wrapper:ident.$ident:ident$($token:tt)*) => {{
+        optional_chain!(@subprop ($wrapper.$ident) ($wrapper.$ident) $($token)*)
+    }};
+
+    // match next identifier as function followed by one or more tokens
+    (@subprop ($msg_acc:expr) ($chain_acc:expr) .$func:ident()$($token:tt)*) => {
+        optional_chain!(@subprop ($msg_acc.$func()) ($chain_acc.and_then(|d| d.$func())) $($token)*)
+    };
+
+    // match next identifier as property followed by one or more tokens
+    (@subprop ($msg_acc:expr) ($chain_acc:expr) .$ident:ident$($token:tt)*) => {
+        optional_chain!(@subprop ($msg_acc.$ident) ($chain_acc.and_then(|d| d.$ident)) $($token)*)
+    };
+
+    // called to close the expression chain and append the message
+    (@subprop ($msg_acc:expr) ($chain_acc:expr)) => {
+        $chain_acc
+            .ok_or(MessageHubProtoError::ExpectedField(stringify!($msg_acc)))
+    };
+}
+
+/// This macro can be used to implement the TryFrom trait for the AstarteType from one or more of
+/// the protobuf types.
+macro_rules! impl_individual_data_to_astarte_type_conversion_traits {
+    (scalar $($typ:ident, $astartedatatype:ident),*; vector $($arraytyp:ident, $astartearraydatatype:ident),*) => {
+        use astarte_message_hub_proto::astarte_data_type_individual::IndividualData;
+
+        impl TryFrom<IndividualData> for AstarteType {
+            type Error = MessageHubProtoError;
+
+            fn try_from(
+                d: IndividualData
+            ) -> Result<Self, Self::Error> {
+
+                match d {
+                    $(
+                    IndividualData::$typ(val) => {
+                        Ok(AstarteType::$astartedatatype(val.into()))
+                    }
+                    )?
+                    IndividualData::AstarteDateTime(val) => {
+                        Ok(AstarteType::DateTime(val.try_into()
+                            .map_err(|e: &str| MessageHubProtoError::DateConversion(e.to_owned()))?))
+                    }
+                    $(
+                    IndividualData::$arraytyp(val) => {
+                        Ok(AstarteType::$astartearraydatatype(val.values.into()))
+                    }
+                    )?
+                    IndividualData::AstarteDateTimeArray(val) => {
+                        let timestamps = val.values.into_iter()
+                            .map(|t| t.try_into())
+                            .collect::<Result<Vec<DateTime<chrono::Utc>>, &str>>()
+                            .map_err(|e| MessageHubProtoError::DateConversion(e.to_owned()))?;
+
+                        Ok(AstarteType::DateTimeArray(timestamps))
+                    }
+                }
+            }
+        }
+    }
+}
+
+impl_individual_data_to_astarte_type_conversion_traits!(
+    scalar
+    AstarteDouble, Double,
+    AstarteInteger,  Integer,
+    AstarteBoolean, Boolean,
+    AstarteLongInteger,LongInteger,
+    AstarteString, String,
+    AstarteBinaryBlob, BinaryBlob;
+    vector
+    AstarteDoubleArray, DoubleArray,
+    AstarteIntegerArray, IntegerArray,
+    AstarteBooleanArray, BooleanArray,
+    AstarteLongIntegerArray, LongIntegerArray,
+    AstarteStringArray, StringArray,
+    AstarteBinaryBlobArray, BinaryBlobArray
+);
+
+impl TryFrom<astarte_message_hub_proto::AstarteDataTypeIndividual> for AstarteType {
+    type Error = MessageHubProtoError;
+
+    fn try_from(
+        value: astarte_message_hub_proto::AstarteDataTypeIndividual,
+    ) -> Result<Self, Self::Error> {
+        optional_chain!(value.individual_data).and_then(TryInto::try_into)
+    }
+}
+
+/// Implements the TryFrom trait for the AstarteDataTypeIndividual for any AstarteType.
+macro_rules! impl_astarte_type_to_individual_data_conversion_traits {
+    ($($typ:ident),*) => {
+        impl TryFrom<AstarteType> for astarte_message_hub_proto::AstarteDataTypeIndividual {
+            type Error = MessageHubProtoError;
+
+            fn try_from(d: AstarteType) -> Result<Self, Self::Error> {
+                match d {
+                    $(
+                    AstarteType::$typ(val) => Ok(val.into()),
+                    )*
+                    AstarteType::Unset => Err(MessageHubProtoError::UnsetConversion),
+                }
+            }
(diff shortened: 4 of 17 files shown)


src/transport/grpc/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+#[derive(Clone)]
+pub struct Grpc {
+    shared: Arc<SharedGrpc>,
+}
+
+impl Grpc {
+    pub(crate) fn new(
+        client: MessageHubClient<tonic::transport::Channel>,
+        stream: tonic::codec::Streaming<AstarteMessage>,
+        uuid: String,
```
harlem88: Change `uuid` String param to [Uuid](https://docs.rs/uuid/latest/uuid/)

``` rust
uuid: Uuid,
```

src/transport/grpc/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+
+    async fn next_event<S>(
+        &self,
+        device: &SharedDevice<S>,
+    ) -> Result<ReceivedEvent<Self::Payload>, crate::Error>
+    where
+        S: PropertyStore,
+    {
+        loop {
+            match self.next_message().await? {
```
harlem88, line 202: The error case on `next_message` should also be handled without propagating it,  for example on device-runtime we handled the tonic status codes `Unavailable` and `Unknown` and in those specific cases we retried attaching the node to the message-hub. https://github.com/edgehog-device-manager/edgehog-device-runtime/blob/a31c8a6c91a08f382ab6411b77104b157ca1b371/src/data/astarte_message_hub_node.rs#L204

harlem88 (changes requested):

src/transport/grpc/convert.rs
```diff
@@ -0,0 +1,1349 @@
...
+use chrono::DateTime;
+
+use crate::transport::grpc::GrpcReceivePayload;
+use crate::transport::ReceivedEvent;
+use crate::types::AstarteType;
+use crate::validate::ValidatedIndividual;
+use crate::validate::ValidatedObject;
+use crate::Aggregation;
+use crate::AstarteDeviceDataEvent;
+use crate::Timestamp;
```
rgallor, line 38: sort lines:

```suggestion
use crate::Aggregation;
use crate::AstarteDeviceDataEvent;
use crate::Timestamp;
use crate::transport::ReceivedEvent;
use crate::transport::grpc::GrpcReceivePayload;
use crate::types::AstarteType;
use crate::validate::ValidatedIndividual;
use crate::validate::ValidatedObject;
```

src/transport/grpc/convert.rs
```diff
@@ -0,0 +1,1349 @@
...
+/// This function can be used to convert a map of (String, crate:types::AstarteType) into a
+/// map of (String,  AstarteDataTypeIndividual).
+pub fn map_values_to_astarte_data_type_individual(
+    value: HashMap<String, AstarteType>,
+) -> Result<
+    HashMap<String, astarte_message_hub_proto::AstarteDataTypeIndividual>,
+    MessageHubProtoError,
+> {
+    let mut map: HashMap<String, astarte_message_hub_proto::AstarteDataTypeIndividual> =
+        Default::default();
```
rgallor, line 383: There is no need for type annotation since you already specified it in the return value of the function. To avoid the warning, use `HashMap::default()` (not `Default::default()`):


```suggestion
    let mut map = HashMap::default();
```

src/transport/grpc/convert.rs
```diff
@@ -0,0 +1,1349 @@
...
+        map.insert(
+            key,
+            astarte_data
+                .individual_data
+                .ok_or(MessageHubProtoError::ConversionError)?
+                .try_into()?,
+        );
+    }
+
+    Ok(map)
```
rgallor, line 408: You could simplify the code by using `try_fold`:


```suggestion
    value.into_iter().try_fold(HashMap::default(), |mut map, (key, astarte_data)|{
        map.insert(
            key,
            astarte_data
                .individual_data
                .ok_or(MessageHubProtoError::ConversionError)?
                .try_into()?,
        );
        Ok(map)
    })
```

src/transport/grpc/convert.rs
```diff
@@ -0,0 +1,1349 @@
...
+) -> Result<
+    HashMap<String, astarte_message_hub_proto::AstarteDataTypeIndividual>,
+    MessageHubProtoError,
+> {
+    let mut map: HashMap<String, astarte_message_hub_proto::AstarteDataTypeIndividual> =
+        Default::default();
+    for (key, astarte_type) in value.into_iter() {
+        map.insert(key, astarte_type.try_into()?);
+    }
+    Ok(map)
```
rgallor, line 387: Alternatively, you could use the `try_fold` method in this way:


```suggestion
    value.into_iter().try_fold(HashMap::default(), |mut map, (key, astarte_type)| {
            map.insert(key, astarte_type.try_into()?);
            Ok(map)
        })
```

The result should be the same.

src/transport/grpc/mod.rs
```diff
@@ -0,0 +1,837 @@
...
+
+    async fn try_detach<S>(
+        client: &mut MessageHubClient<tonic::transport::Channel>,
+        uuid: S,
+    ) -> Result<(), GrpcTransportError>
+    where
+        S: ToString,
+    {
+        // During the detach phase only the uuid is needed we can pass an empty array
+        // as the interface_json since the interfaces are alredy known to the message hub
```
rgallor, line 136: typo `already`

src/transport/grpc/mod.rs
```diff
@@ -0,0 +1,837 @@
...
+/*
+/// Implemented to correctly detach the device from the message hub
+///
+/// This implementation currently blocks the current thread to wait
+/// for the complete disconnection from the message hub.
+impl Drop for Grpc {
+    fn drop(&mut self) {
+    }
+}
+*/
```
rgallor, line 311: Is this snippet still necessary?

rgallor (changes requested):

src/validate.rs
```diff
@@ -100,56 +100,60 @@ impl<'a> ValidatedIndividual<'a> {
...
-                });
+            ref data => {
+                if *data != mapping.mapping_type() {
+                    return Err(UserValidationError::SerializeType {
+                        expected: mapping.mapping_type().to_string(),
+                        got: data.display_type().to_string(),
+                    });
+                }
             }
         }
```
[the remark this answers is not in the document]
joshuachp: The suggestion is incorrect and introduces a bug where the unset property is always different from the mapping type

Cargo.toml
```diff
@@ -119,6 +127,9 @@ syn = "2.0.0"
 tempfile = "3.6.0"
 thiserror = "1.0.35"
 tokio = "1.18.0"
+tokio-stream = "0.1.14"
+tonic = "0.10.2"
+tower = "0.4.13"
```
joshuachp, line 132: I wonder if we can lower the minimum required version to 0 for these dependencies.

joshuachp (changes requested):

src/transport/grpc/mod.rs
```diff
@@ -0,0 +1,837 @@
...
+    types::AstarteType,
+    validate::{ValidatedIndividual, ValidatedObject},
+    Interface, Timestamp,
+};
+
+use super::{Disconnect, Publish, Receive, ReceivedEvent, Register};
+
+use self::convert::map_values_to_astarte_type;
+
+/// Errors raised while using the [`Grpc`] transport
```
joshuachp, line 56:
```suggestion
/// Errors raised while using the [`Grpc`] transport
#[non_exhaustive]
```

src/transport/grpc/mod.rs
```diff
@@ -0,0 +1,837 @@
...
+    async fn next_message(&self) -> Result<Option<AstarteMessage>, GrpcTransportError> {
+        self.stream
+            .lock()
+            .await
+            .message()
+            .await
+            .map_err(GrpcTransportError::from)
+    }
+
+    async fn try_attach(
```
joshuachp, line 117: I would remove the try, since there is no version which doesn't return an error

src/transport/grpc/mod.rs
```diff
@@ -0,0 +1,837 @@
...
+/// This implementation currently blocks the current thread to wait
+/// for the complete disconnection from the message hub.
+impl Drop for Grpc {
+    fn drop(&mut self) {
+    }
+}
+*/
+
+/// Internal struct holding the received grpc message
+#[derive(Debug, PartialEq)]
```
joshuachp, line 314:
```suggestion
#[derive(Debug, Clone, PartialEq)]
```

src/transport/grpc/mod.rs
```diff
@@ -0,0 +1,837 @@
...
+    client: Mutex<MessageHubClient<tonic::transport::Channel>>,
+    stream: Mutex<tonic::codec::Streaming<AstarteMessage>>,
+    uuid: Uuid,
+}
+
+/// This struct represents a GRPC connection handler for an Astarte device. It manages the
+/// interaction with the [astarte-message-hub](https://github.com/astarte-platform/astarte-message-hub), sending and receiving [`AstarteMessage`]
+/// following the Astarte message hub protocol.
+#[derive(Clone)]
+pub struct Grpc {
```
joshuachp, line 85: I whould add a manual impl of debug on the inner shared fileds, or at least of the struct name, without locking

src/transport/grpc/mod.rs
```diff
@@ -0,0 +1,837 @@
...
+    timestamp: Option<Timestamp>,
+}
+
+impl GrpcReceivePayload {
+    pub(crate) fn new(data: astarte_data_type::Data, timestamp: Option<Timestamp>) -> Self {
+        Self { data, timestamp }
+    }
+}
+
+pub struct GrpcConfig {
```
joshuachp, line 326:
```suggestion
#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
pub struct GrpcConfig {
```

joshuachp: And probably display since it's bug, but not really necessary

src/transport/mod.rs
```diff
@@ -44,10 +44,13 @@ use crate::{
 };
 
 pub mod mqtt;
+//#[cfg(feature="message-hub")]
+pub mod grpc;
 
 /// Holds generic event data such as interface name and path
 /// The payload must be deserialized after verification with the
 /// specific [`Connection::deserialize_individual`] or [`Connection::serialize_individual`]
+#[derive(Debug)]
```
joshuachp, line 53:
```suggestion
#[derive(Debug, Clone)]
```

src/transport/mqtt/mod.rs
```diff
@@ -562,6 +562,8 @@ pub enum MqttConnectionError {
     Pairing(#[from] PairingError),
     #[error(transparent)]
     Pki(#[from] webpki::Error),
+    #[error("Error while loading session data to perform the mqtt connection: {0}")]
+    PropLoad(#[from] StoreError),
 }
```
joshuachp, line 567: The error should be non_exauhustive

joshuachp (changes requested):

Cargo.toml
```diff
@@ -56,6 +57,7 @@ log = { workspace = true }
 once_cell = { workspace = true }
 openssl = { workspace = true, optional = true }
 p384 = { workspace = true }
+pbjson-types = { workspace = true }
```
harlem88, line 60: Remove this crate, you can use the astarte-message-hub-proto re-export.
```suggestion
```

Cargo.toml
```diff
@@ -67,6 +69,7 @@ serde_json = { workspace = true }
 sqlx = { workspace = true, features = ["runtime-tokio", "sqlite", "macros", "migrate"] }
 thiserror = { workspace = true }
 tokio = { workspace = true, features = ["parking_lot", "macros"] }
+tonic = { workspace = true } # , optional = true
```
harlem88, line 72: Remove this crate, you can use the astarte-message-hub-proto re-export.
```suggestion
```

Cargo.toml
```diff
@@ -103,6 +110,7 @@ mockall = "0.11.4"
 once_cell = "1.18.0"
 openssl = "0.10.46"
 p384 = "0.13.0"
+pbjson-types = "0.6.0"
```
harlem88, line 113:
```suggestion
```

Cargo.toml
```diff
@@ -119,6 +127,9 @@ syn = "2.0.0"
 tempfile = "3.6.0"
 thiserror = "1.0.35"
 tokio = "1.18.0"
+tokio-stream = "0.1.14"
+tonic = "0.10.2"
```
harlem88, line 131:
```suggestion
```

harlem88 (changes requested):

src/lib.rs
```diff
@@ -744,18 +747,24 @@ where
...
+                    device
+                        .handle_event(&interface, &path, payload)
+                        .await
+                        .map(|aggregation| AstarteDeviceDataEvent {
+                            interface,
+                            path,
+                            data: aggregation,
+                        });
 
                 device.tx.send(data).await.expect("Channel dropped")
```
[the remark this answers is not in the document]
lucaato: The error on a sens of a `mpsc` channel does mean that the other end of the channel was dropped so we should probably exit this loop ? Or do you think that just logging is fine ?

src/transport/grpc/mod.rs
lucaato: Nope, snippet removed.

src/transport/grpc/convert.rs
```diff
@@ -0,0 +1,1349 @@
...
+    #[error(transparent)]
+    TryFromIntError(#[from] TryFromIntError),
+
+    /// Failed to convert between types
+    #[error("unable to convert type")]
+    ConversionError,
+
+    /// Expected field was not found
+    #[error("Missing the expected field '{0}'")]
+    ExpectedField(String),
```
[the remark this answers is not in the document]
lucaato: Yeah, nice catch.

src/error.rs
```diff
@@ -115,4 +116,13 @@ pub enum Error {
...
     MqttConnection(#[from] MqttConnectionError),
+
+    // TODO #[cfg(feature="grpc")]
+    #[error(transparent)]
+    GrpcTransport(#[from] GrpcTransportError),
+
+    // TODO #[cfg(feature="grpc")]
+    // FIXME could be included in the GrpcTransport error
+    #[error(transparent)]
+    MessageHubProto(#[from] MessageHubProtoError),
```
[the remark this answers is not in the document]
lucaato: I moved the `MessageHubProto` error as a variant of the `GrpcTransportError` enum to still keep conversion errors separated from transport errors. Let me know what you think.

src/transport/grpc/mod.rs
lucaato: I've changed the error handling code to take into account what you wrote, let me know if this works for you. It's still a bit different from the code you've linked but I think it makes sense to handle the `Ok(None)` case since the [documentation](https://docs.rs/tonic/latest/tonic/codec/struct.Streaming.html#method.message) declares that no more messages will be available.

src/transport/grpc/convert.rs
lucaato: Refactored the conversion.

src/transport/grpc/convert.rs
lucaato: Refactored the conversion to use `collect` instead to have a pre computed `HashMap` size.

src/transport/grpc/convert.rs
```diff
@@ -0,0 +1,1349 @@
...
+
+#[derive(thiserror::Error, Debug)]
+pub enum MessageHubProtoError {
+    /// Wrapper for integer conversion errors
+    #[error(transparent)]
+    TryFromIntError(#[from] TryFromIntError),
+
+    /// Failed to convert between types
+    #[error("unable to convert type")]
+    ConversionError,
```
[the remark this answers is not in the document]
lucaato: I've refactored the code to use other error types instead and it is now way better.

src/transport/grpc/mod.rs
lucaato: Why not deriving the impl ? The derived impl should not lock anyways. And i will also derive `Debug` for the `SharedGrpc` object.

src/lib.rs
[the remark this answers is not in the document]
joshuachp: Just logging should be fine

src/transport/grpc/convert.rs
lucaato: Changed the impl to use iterators.

Cargo.toml
lucaato: Lowered `tokio-stream` and removed the other two since those are now imported from the `astarte-message-hub-proto` crate.

src/lib.rs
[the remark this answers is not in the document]
lucaato: Logged the error instead of panicking.

src/transport/grpc/convert.rs
```diff
@@ -0,0 +1,1349 @@
...
+    /// Date conversion error
+    #[error("Error while converting a proto date: {0}")]
+    DateConversion(String),
+}
+
+// TODO could be rewrote to accept chaining of accesses
+macro_rules! expect_field {
+    ($wrapper:ident.$field:ident) => {{
+        $wrapper.$field.ok_or(MessageHubProtoError::ExpectedField(
+            stringify!($field).to_string(),
```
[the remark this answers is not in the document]
lucaato: Resolved as suggested.

src/transport/grpc/mod.rs
lucaato: It is actually pretty noisy
```
Grpc { shared: SharedGrpc { client: Mutex { data: MessageHubClient { inner: Grpc { inner: Channel, origin: /, compression_encoding: None, accept_compression_encodings: EnabledCompressionEncodings, max_decoding_message_size: None, max_encoding_message_size: None } } }, stream: Mutex { data: Streaming }, uuid: 67e55044-10b1-426f-9247-bb680e5fe0c8 } }
```
For now i just print `Grpc { uuid: 67e55044-10b1-426f-9247-bb680e5fe0c8, .. }`

src/transport/grpc/mod.rs
lucaato: Avoided making it displayable but I derived the other traits like you suggested.

Cargo.toml
lucaato: Used the re exposed traits

src/transport/grpc/convert.rs
```diff
@@ -0,0 +1,1347 @@
...
+    /// Cannot convert [`AstarteType::Unset`] to a proto [`IndividualData`]
+    #[error("Cannot perform conversion of an Unset to a proto IndividualData")]
+    UnsetConversion,
+
+    /// Date conversion error
+    #[error("Error while converting a proto date: {0}")]
+    DateConversion(String),
+}
+
+macro_rules! optional_chain {
```
rgallor, line 58: I would add a description about how this macro works

rgallor (changes requested):

src/transport/grpc/convert.rs
```diff
@@ -0,0 +1,1347 @@
...
+        crate::Interface::from_str(interface_str.as_ref()).map_err(Self::Error::Interface)
+    }
+}
+
+// FIXME this is probably useless ?
+impl TryFrom<AstarteDeviceDataEvent> for astarte_message_hub_proto::AstarteMessage {
+    type Error = MessageHubProtoError;
+
+    fn try_from(value: AstarteDeviceDataEvent) -> Result<Self, Self::Error> {
+        use astarte_message_hub_proto::astarte_message::Payload;
```
rgallor, line 245: It seems you are using `Payload` several times. Why don't you import it directly on the top of this file?

src/transport/grpc/convert.rs
```diff
@@ -0,0 +1,1347 @@
...
+    fn try_from(
+        interface: astarte_message_hub_proto::types::InterfaceJson,
+    ) -> Result<Self, Self::Error> {
+        let interface_str = String::from_utf8_lossy(&interface.0);
+
+        crate::Interface::from_str(interface_str.as_ref()).map_err(Self::Error::Interface)
+    }
+}
+
+// FIXME this is probably useless ?
```
rgallor, line 240: Is this comment still necessary?

src/transport/grpc/convert.rs
```diff
@@ -0,0 +1,1347 @@
...
+    fn try_from(value: Aggregation) -> Result<Self, Self::Error> {
+        use astarte_message_hub_proto::astarte_data_type::Data;
+        use astarte_message_hub_proto::astarte_message::Payload;
+
+        let payload = match value {
+            Aggregation::Individual(astarte_type) => {
+                if let AstarteType::Unset = astarte_type {
+                    Payload::AstarteUnset(astarte_message_hub_proto::AstarteUnset {})
+                } else {
+                    let individual_type: astarte_message_hub_proto::AstarteDataTypeIndividual =
```
rgallor, line 338: You can avoid specifying the type

src/transport/grpc/convert.rs
```diff
@@ -0,0 +1,1347 @@
...
+impl<'a> TryFrom<ValidatedObject<'a>> for astarte_message_hub_proto::AstarteMessage {
+    type Error = MessageHubProtoError;
+
+    fn try_from(value: ValidatedObject<'a>) -> Result<Self, Self::Error> {
+        use astarte_message_hub_proto::astarte_message::Payload;
+
+        let interface_name = value.object().interface.interface_name().to_owned();
+        let path = value.path().as_str().to_owned();
+        let timestamp: Option<pbjson_types::Timestamp> = value.timestamp().map(|t| t.into());
+        let astarte_data: astarte_message_hub_proto::AstarteDataType = value.into_data()
```
rgallor, line 309: You can avoid specifying the type

src/transport/grpc/convert.rs
```diff
@@ -0,0 +1,1347 @@
...
+                    let individual_type: astarte_message_hub_proto::AstarteDataTypeIndividual =
+                        astarte_type.try_into()?;
+
+                    Payload::AstarteData(astarte_message_hub_proto::AstarteDataType {
+                        data: Some(Data::AstarteIndividual(individual_type)),
+                    })
+                }
+            }
+            Aggregation::Object(astarte_map) => {
+                let astarte_data: astarte_message_hub_proto::AstarteDataType = astarte_map
```
rgallor, line 347: You can avoid specifying the type

src/transport/grpc/convert.rs
```diff
@@ -0,0 +1,1347 @@
...
+
+impl TryFrom<AstarteType> for astarte_message_hub_proto::astarte_message::Payload {
+    type Error = MessageHubProtoError;
+
+    fn try_from(astarte_device_sdk_type: AstarteType) -> Result<Self, Self::Error> {
+        use astarte_message_hub_proto::astarte_data_type::Data;
+        use astarte_message_hub_proto::astarte_message::Payload;
+        use astarte_message_hub_proto::AstarteDataType;
+
+        let payload = if let AstarteType::Unset = astarte_device_sdk_type {
```
rgallor, line 369: Here you could have used a match rather than if-let-else

rgallor:
```rust
let payload = match astarte_device_sdk_type {
            AstarteType::Unset => Payload::AstarteUnset(astarte_message_hub_proto::AstarteUnset {}),
            _ => astarte_message_hub_proto::AstarteDataTypeIndividual::try_from(
                astarte_device_sdk_type,
            )
            .map(|individual_type| {
                Payload::AstarteData(AstarteDataType {
                    data: Some(Data::AstarteIndividual(individual_type)),
                })
            })?,
        };
```

src/transport/grpc/convert.rs
```diff
@@ -0,0 +1,1347 @@
...
+    use super::map_values_to_astarte_data_type_individual;
+    use super::AstarteType;
+    use super::IndividualData;
+    use super::MessageHubProtoError;
+
+    #[test]
+    fn proto_astarte_double_into_astarte_device_sdk_type_success() {
+        let value: f64 = 15.5;
+        let expected_double_value = IndividualData::AstarteDouble(value);
+        let astarte_type: AstarteType = expected_double_value.try_into().unwrap();
```
rgallor, line 433: type definition not necessary

rgallor: same in the following tests for all astarte_type variables

rgallor (changes requested):

src/transport/grpc/mod.rs
```diff
@@ -0,0 +1,865 @@
...
+                stream: Mutex::new(stream),
+                uuid,
+            }),
+        }
+    }
+
+    /// Polls a message from the tonic stream and tries reattaching if necessary
+    ///
+    /// An [`Option`] is returned directly from the [`tonic::codec::Streaming::message`] method.
+    /// A result of [`Option::None`] signals a disconnectiond and should be handled by the caller
```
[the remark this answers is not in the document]
rgallor: also you could write [`None`] instead of [`Option::None`]

src/transport/grpc/mod.rs
```diff
@@ -0,0 +1,865 @@
...
+            >,
+        >,
+        /// This stream contains requests received by the server
+        server_received: mpsc::Sender<ServerReceivedRequest>,
+    }
+
+    impl TestMessageHubServer {
+        fn new(
+            server_send: mpsc::Receiver<
+                Result<astarte_message_hub_proto::AstarteMessage, tonic::Status>,
```
rgallor, line 456: path prefix not necessary: astarte_message_hub_proto

src/transport/grpc/mod.rs
```diff
@@ -0,0 +1,865 @@
...
+            Self {
+                server_send: Mutex::new(Some(server_send)),
+                server_received,
+            }
+        }
+    }
+
+    #[async_trait]
+    impl MessageHub for TestMessageHubServer {
+        type AttachStream = tokio_stream::wrappers::ReceiverStream<
```
rgallor, line 469: path prefix not necessary: tokio_stream::wrappers::

src/transport/grpc/mod.rs
```diff
@@ -0,0 +1,865 @@
...
+            Result<astarte_message_hub_proto::AstarteMessage, tonic::Status>,
```
rgallor, line 470: path prefix not necessary: astarte_message_hub_proto

src/transport/grpc/mod.rs
```diff
@@ -0,0 +1,865 @@
...
+        >;
+
+        async fn attach(
+            &self,
+            request: tonic::Request<super::Node>,
+        ) -> std::result::Result<tonic::Response<Self::AttachStream>, tonic::Status> {
```
rgallor, line 476: path prefix not necessary: std::result::

src/transport/grpc/mod.rs
```diff
@@ -0,0 +1,865 @@
...
+
+    #[async_trait]
+    impl MessageHub for TestMessageHubServer {
+        type AttachStream = tokio_stream::wrappers::ReceiverStream<
+            Result<astarte_message_hub_proto::AstarteMessage, tonic::Status>,
+        >;
+
+        async fn attach(
+            &self,
+            request: tonic::Request<super::Node>,
```
rgallor, line 475: 'super' not necessary

rgallor: same in the following function

src/transport/grpc/mod.rs
rgallor: same in the following function

src/transport/grpc/mod.rs
```diff
@@ -0,0 +1,865 @@
...
+            Ok(tonic::Response::new(pbjson_types::Empty::default()))
+        }
+    }
+
+    fn make_server(
+        sock: TcpListener,
+        server: TestMessageHubServer,
+    ) -> Result<impl Future<Output = ()>, Box<dyn std::error::Error>> {
+        //let uds = UnixListener::bind(socket)?;
+        //let stream = UnixListenerStream::new(uds);
```
rgallor, line 518: are theese comments still necessary?

src/transport/grpc/mod.rs
```diff
@@ -0,0 +1,865 @@
...
+    ) -> Result<Grpc, Box<dyn std::error::Error>> {
+        let node_data = NodeData::try_from_unlocked(ID, interfaces)?;
+        let stream = Grpc::attach(&mut message_hub_client, node_data).await?;
+
+        Ok(Grpc::new(message_hub_client, stream, ID))
+    }
+
+    struct TestServerChannels {
+        server_response_sender:
+            mpsc::Sender<Result<astarte_message_hub_proto::AstarteMessage, tonic::Status>>,
```
rgallor, line 585: path prefix not necessary: astarte_message_hub_proto::

rgallor (changes requested):

src/transport/grpc/convert.rs
lucaato: There are many similar names between the structs defined in the sdk and the one generated from the protobuf files. I think it has been done to make it immediately clear to which package the type is referring to and I did not change it, if you think it would be better it can be done, for me it works either way.

src/transport/grpc/convert.rs
lucaato: Removed

src/transport/grpc/convert.rs
rgallor: To me is also fine. Btw, you could think of renaming `Payload` so that there is no confusion (e.g., `ProtoPayload`, `MsgHubPayload` or similar)

src/transport/grpc/convert.rs
```diff
@@ -0,0 +1,1347 @@
...
+use chrono::DateTime;
+
+use crate::{
+    transport::grpc::GrpcReceivePayload, transport::ReceivedEvent, types::AstarteType,
+    validate::ValidatedIndividual, validate::ValidatedObject, Aggregation, AstarteDeviceDataEvent,
+    Timestamp,
+};
+
+#[non_exhaustive]
+#[derive(thiserror::Error, Debug)]
```
joshuachp, line 39:
```suggestion

/// Error returned by the Message Hub.
#[non_exhaustive]
#[derive(thiserror::Error, Debug)]
```

src/transport/grpc/convert.rs
```diff
@@ -0,0 +1,1347 @@
...
+    }
+}
+
+impl TryFrom<astarte_message_hub_proto::types::InterfaceJson> for crate::Interface {
+    type Error = crate::Error;
+
+    fn try_from(
+        interface: astarte_message_hub_proto::types::InterfaceJson,
+    ) -> Result<Self, Self::Error> {
+        let interface_str = String::from_utf8_lossy(&interface.0);
```
joshuachp, line 244: This should be an hard error, you can use `str::from_utf8`

src/transport/grpc/convert.rs
```diff
@@ -0,0 +1,1353 @@
...
+
+impl TryFrom<astarte_message_hub_proto::AstarteDataTypeIndividual> for AstarteType {
+    type Error = MessageHubProtoError;
+
+    fn try_from(
+        value: astarte_message_hub_proto::AstarteDataTypeIndividual,
+    ) -> Result<Self, Self::Error> {
+        let individual_data = optional_chain!(value.individual_data)?;
+
+        individual_data.try_into()
```
joshuachp, line 166:
```suggestion
        optional_chain!(value.individual_data).and_then(TryInto::try_into)
```

joshuachp (changes requested):

src/transport/grpc/mod.rs
```diff
@@ -0,0 +1,1008 @@
...
+    }
+
+    async fn reattach(&self, data: NodeData) -> Result<(), GrpcTransportError> {
+        // the lock on stream is actually intended since we are detaching and re-attaching
+        // we want to make sure no one uses the stream while the client is detached
+        let mut stream = self.stream.lock().await;
+        let mut client = self.client.lock().await;
+
+        client
+            .detach(Node::new(self.uuid, &[] as &[Vec<u8>]))
```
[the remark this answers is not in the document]
joshuachp: Or `&Vec::<u8>::new()` if you need the type

src/transport/grpc/mod.rs
```diff
@@ -0,0 +1,1008 @@
...
+}
+
+#[async_trait]
+impl Disconnect for Grpc {
+    async fn disconnect(mut self) -> Result<(), crate::Error> {
+        if let Some(SharedGrpc {
+            client,
+            stream: _stream,
+            uuid,
+        }) = Arc::get_mut(&mut self.shared)
```
joshuachp, line 326:
```suggestion
        let SharedGrpc {
            client,
            stream: _stream,
            uuid,
        }) = Arc::get_mut(&mut self.shared).ok_or(..)?;
```

src/transport/grpc/convert.rs
lucaato: I did this in the end since I liked it better.

src/transport/grpc/convert.rs
lucaato: That's correct.

Cargo.toml
```diff
@@ -79,14 +80,17 @@ env_logger = { workspace = true }
...
 
 [features]
 derive = ["dep:astarte-device-sdk-derive"]
 openssl = ["dep:openssl"]
+message-hub = ["dep:astarte-message-hub-proto"]
 
 [workspace.dependencies]
 astarte-device-sdk-derive = { version = "0.6.1", path = "./astarte-device-sdk-derive" }
-async-trait = "0.1.41"
+astarte-message-hub-proto = "0.6.0"
```
harlem88, line 92:
```suggestion
astarte-message-hub-proto = "0.6.1"
```

harlem88 (changes requested):

src/transport/grpc/mod.rs
```diff
@@ -0,0 +1,1003 @@
...
+
+                    let mut stream = self.stream.lock().await;
+                    let mut client = self.client.lock().await;
+
+                    *stream = Grpc::attach(&mut client, data).await?;
+                }
+                Err(s) => {
+                    return Err(GrpcTransportError::from(s).into());
+                }
+            }
```
harlem88, line 248: What do you think to merge `Ok(none)` and `Err(s) Unavailable & Unknown` statement? Something like that
```suggestion
                Err(s)
                    if s.code() != tonic::Code::Unavailable && s.code() != tonic::Code::Unknown =>
                {
                    return Err(GrpcTransportError::from(s).into());
                }
                _ => {
                   // try reattaching

                    let data =
                        NodeData::try_from_unlocked(self.uuid, &*device.interfaces.read().await)?;

                    let mut stream = self.stream.lock().await;
                    let mut client = self.client.lock().await;

                    *stream = Grpc::attach(&mut client, data).await?;
                }
            }
```

harlem88 (changes requested):

src/transport/grpc/mod.rs
lucaato: Yes, that's better.

harlem88 (approved):

rgallor (approved):

joshuachp (approved):

PR #281, opened 2024-02-06
feat(metadata): add node id into gRPC metadata

rgallor:


src/transport/grpc/mod.rs
```diff
@@ -73,11 +79,28 @@ pub enum GrpcTransportError {
...
+
+impl NodeIdInterceptor {
+    pub(crate) fn new(uuid: Uuid) -> Self {
+        Self(uuid)
+    }
+}
+
+impl Interceptor for NodeIdInterceptor {
+    fn call(&mut self, mut req: Request<()>) -> Result<Request<()>, Status> {
+        let val = MetadataValue::from_str(&self.0.to_string())
```
lucaato, line 92: Wouldnt it be possible to perform the conversion to a `MetadataValue` just one time in the `new` method ? this would mean that every interceptor call would just insert into the request our metadata.

src/transport/grpc/mod.rs
```diff
@@ -73,11 +79,28 @@ pub enum GrpcTransportError {
...
+            .map_err(|_| Status::new(Code::InvalidArgument, "invalid metadata value"))?;
+        req.metadata_mut().insert("node_id", val);
```
lucaato, line 94: I would move this `node_id` string to a constant in the impl block of `NodeIdInterceptor`

lucaato (changes requested):
Let me know if the comments make sense to you

src/transport/grpc/mod.rs
rgallor: It could be possible but I see 2 drawbacks:
1. the `new` method should return a `Result` (which is not a big deal)
2. the `call` method has a mut reference to self and the `insert` method requires ownership of the `MetadataValue`, therefore we should clone the value at every insertion (which should approximately have the same overhead of creating a new `MetadataValue` each time).

Btw, your observation makes sense. Let's see what the others think about it

src/transport/grpc/mod.rs
```diff
@@ -73,11 +79,30 @@ pub enum GrpcTransportError {
...
+        Ok(req)
+    }
+}
+
 /// Shared data of the grpc connection, this struct is internal to the [`Grpc`] connection
 /// where is wrapped in an arc to share an immutable reference across tasks.
 #[derive(Debug)]
 pub struct SharedGrpc {
-    client: Mutex<MessageHubClient<tonic::transport::Channel>>,
+    client: Mutex<MessageHubClient<InterceptedService<Channel, NodeIdInterceptor>>>,
```
joshuachp, line 105: Add a type alias to simplify the message hub client, in all the places it's used.

joshuachp (changes requested):
It's stranget that you had to downgrade itertools in the cargo.lock. I would double check that

harlem88 (changes requested):

harlem88 (changes requested):
Add node-id changes to the Changelog file

src/transport/grpc/mod.rs
```diff
@@ -124,9 +151,9 @@ impl Grpc {
             .map_err(GrpcTransportError::from)
     }
 
-    // TODO this should be consuming the client but Arc::into_inner is not stsable in version 1.66.1
+    // TODO this should be consuming the client but Arc::into_inner is not stable in version 1.66.1
```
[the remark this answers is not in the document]
rgallor: I updated the code using `into_inner`

src/transport/grpc/mod.rs
```diff
@@ -309,12 +336,10 @@ impl Disconnect for Grpc {
...
             stream: _stream,
             uuid,
-        } = Arc::get_mut(&mut self.shared)
+        } = Arc::into_inner(self.shared)
             .ok_or::<crate::Error>(GrpcTransportError::GracefulClose.into())?;
 
-        Self::detach(client.get_mut(), uuid)
-            .await
-            .map_err(|e| e.into())
+        Self::detach(client, &uuid).await.map_err(|e| e.into())
```
joshuachp, line 342: This should be:

```rust
        let SharedGrpc {
            client,
            stream: _stream,
            uuid,
        } = Arc::into_inner(self.shared)
            .ok_or(GrpcTransportError::GracefulClose)?;
        let client = client.into_inner()
               .map_err(|err| {
                  error!("client {err}");

                  GrpcTransportError::GracefulClose
                })?;


        Self::detach(client, &uuid).await.map_err(|e| e.into())
```

joshuachp (changes requested):

src/transport/grpc/mod.rs
```diff
@@ -124,14 +151,14 @@ impl Grpc {
             .map_err(GrpcTransportError::from)
     }
 
-    // TODO this should be consuming the client but Arc::into_inner is not stsable in version 1.66.1
     async fn detach(
-        client: &mut MessageHubClient<tonic::transport::Channel>,
+        client: Mutex<MessageHubClientWithInterceptor>,
```
joshuachp, line 155:
```suggestion
        client: mut MessageHubClientWithInterceptor,
```

src/transport/grpc/mod.rs
```diff
@@ -124,14 +151,14 @@ impl Grpc {
...
         uuid: &Uuid,
     ) -> Result<(), GrpcTransportError> {
         // During the detach phase only the uuid is needed we can pass an empty array
         // as the interface_json since the interfaces are already known to the message hub
         // this api will change in the future
+        let mut client = client.lock().await;
```
joshuachp, line 161:
```suggestion
```

joshuachp (changes requested):

src/transport/grpc/mod.rs
```diff
@@ -411,7 +442,10 @@ mod test {
 
     use super::super::{test::mock_shared_device, Publish, Receive, ReceivedEvent};
 
-    use super::{Grpc, GrpcReceivePayload, GrpcTransportError, NodeData};
+    use super::{
+        Grpc, GrpcReceivePayload, GrpcTransportError, MessageHubClientWithInterceptor, NodeData,
+        NodeIdInterceptor,
+    };
```
joshuachp, line 448: Also for the use super before this can be collapsed in

```suggestion
    use super::*;
```

joshuachp (changes requested):

harlem88 (approved):

joshuachp (approved):
LGTM

harlem88 (approved):

PR #293, opened 2024-03-14
feat(interfaces): add many interfaces at once

### LINKED ISSUE #291: Implement the ability to add multiple interfaces
Currently, to add multiple interfaces you have to call `add_interface` for each one.

This has the disadvantage that the introspection is re-sent for each one.

We need to introduce a method to add multiple interfaces at once.

joshuachp:
Add the possibility to add multiple interfaces all at once. For the MQTT transport we send a single introspection and subscribe to many topics at once. Instead, for the gRPC we still need to disconnect and reattach the node.

Closes #291

### DIFF
--- a/src/interfaces.rs
+++ b/src/interfaces.rs
@@ -24,7 +24,7 @@ use std::fmt::Display;
 use std::ops::Deref;
 
 use itertools::Itertools;
-use log::debug;
+use log::{debug, trace};
 
 use crate::{
     interface::{
@@ -65,7 +65,7 @@ impl Interfaces {
     ) -> Result<Option<Validated>, InterfaceError> {
         match self.interfaces.get(interface.interface_name()) {
             Some(prev) => {
-                debug!(
+                trace!(
                     "Interface {} already present, validating new version",
                     interface.interface_name()
                 );
@@ -80,7 +80,7 @@ impl Interfaces {
                 interface.validate_with(prev)?;
             }
             None => {
-                debug!("Interface {} not present", interface.interface_name());
+                trace!("Interface {} not present", interface.interface_name());
             }
         }
 
--- a/e2e-test/src/main.rs
+++ b/e2e-test/src/main.rs
@@ -163,6 +163,7 @@ async fn main() -> eyre::Result<()> {
 
     // Add the remaining interfaces
     let additional_interfaces = read_additional_interfaces()?;
+    debug!("adding {} interfaces", additional_interfaces.len());
     device_cpy.extend_interfaces(additional_interfaces).await?;
 
     tasks.spawn(async move {
--- a/src/interface/mapping/endpoint.rs
+++ b/src/interface/mapping/endpoint.rs
@@ -23,7 +23,7 @@ use std::{
 };
 
 use itertools::{EitherOrBoth, Itertools};
-use log::{debug, error, info, trace};
+use log::{error, trace};
 
 use super::path::MappingPath;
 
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `AstarteDeviceSdk`.
 - Trait `FromEvent` to convert a generic object aggregate into a Rust struct.
 - Implementation of the connection over GRPC to the message hub.
+- Add many interfaces at once with `extend_interfaces`
+  [#293](https://github.com/astarte-platform/astarte-device-sdk-rust/pull/293)
```
harlem88, line 28: Move this change to  `## [Unreleased]` tag

harlem88 (changes requested):

CHANGELOG.md
joshuachp: Done

src/interfaces.rs
```diff
@@ -50,48 +51,48 @@ impl Interfaces {
...
-        let prev = match entry {
-            Entry::Occupied(mut entry) => {
+    /// Validate that and interface can be added.
+    ///
+    /// It will return [`None`] if the interface is already present.
+    pub(crate) fn validate(
+        &self,
+        interface: Interface,
+    ) -> Result<Option<Validated>, InterfaceError> {
+        match self.interfaces.get(interface.interface_name()) {
```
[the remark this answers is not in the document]
joshuachp: This cannot be converted in a let else

CHANGELOG.md
```diff
@@ -5,8 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-### Added
 - Introduce Node ID into gRPC metadata.
+- Add one or more interfaces at once with `extend_interfaces`
```
harlem88, line 9: `Added` was removed
```suggestion
### Added
- Introduce Node ID into gRPC metadata.
- Add one or more interfaces at once with `extend_interfaces`
```

harlem88 (changes requested):

CHANGELOG.md
joshuachp: Re-added

harlem88 (approved):

src/introspection.rs
```diff
@@ -0,0 +1,323 @@
...
+        debug!("Interfaces added");
+
+        Ok(())
+    }
+
+    async fn add_interface_from_file<P>(&self, file_path: P) -> Result<(), Error>
+    where
+        P: AsRef<Path> + Send,
+    {
+        let interface = fs::read_to_string(&file_path).map_err(|err| AddInterfaceError::Io {
```
rgallor, line 168: why didn't you use `tokio::fs` instead of `std::fs`?

rgallor (changes requested):

src/introspection.rs
rgallor: Also, there is another place where `std::fs` is used, you could fix also that one

joshuachp: Because the existing function call was synchronous and there was no need to change the API. Changing this should be done in another issue, with its own considerations.

The builder is used at the start of the program, so blocking the thread should be not be a problem, and having a sync call is more ergonomic without all the awaits. Maybe we should provide both APIs. Either way, I'm opening a new issue about this.

rgallor (approved):

PR #296, opened 2024-03-20
refactor(unset): remove the AstarteType::Unset

joshuachp:
Move the Unset to the `Aggregation` enum and introduce a new `unset` method. This simplify the checks since you can only unset a property and the AstarteType::Unset is not needed in the others code paths. Also this simplifies the conversions in the gRPC code that now in infallible.

### DIFF
--- a/e2e-test/src/main.rs
+++ b/e2e-test/src/main.rs
@@ -215,15 +215,14 @@ async fn main() -> eyre::Result<()> {
             } else if event.interface == test_cfg.interface_property_so {
                 let mut rx_data = rx_data_ind_prop.lock().await;
 
-                let mut path = event.path.clone();
-                path.remove(0); // Remove first forward slash
+                let path = event.path.as_str();
                 let (sensor_n, key) = path
-                    .split_once('/')
-                    .ok_or_else(|| eyre!("Incorrect path in message {:?}", event))?;
+                    .strip_prefix('/')
+                    .and_then(|s| s.split_once('/'))
+                    .unwrap_or_else(|| panic!("Incorrect path in message {:?}", event.clone()));
 
                 rx_data.0 = sensor_n.to_string();
-
-                match event.clone().data {
+                match event.data {
                     Value::Individual(var) => {
                         rx_data.1.insert(key.to_string(), var);
                     }
--- a/e2e-test/src/main.rs
+++ b/e2e-test/src/main.rs
@@ -29,15 +29,16 @@ use std::collections::HashMap;
 use std::env;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
-use std::sync::{Arc, Mutex};
+use std::sync::Arc;
 use std::time::Duration;
 
 use astarte_device_sdk::{Interface, Value};
-use eyre::WrapErr;
+use eyre::{eyre, OptionExt, WrapErr};
 use itertools::Itertools;
 use log::{debug, error, info};
 use reqwest::StatusCode;
 use serde_json::Value as JsonValue;
+use tokio::sync::Mutex;
 use tokio::task::JoinSet;
 use tokio::time;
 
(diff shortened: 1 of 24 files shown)


src/interface/mod.rs
```diff
@@ -183,7 +183,10 @@ impl Interface {
     }
 
     /// Get a [`MappingRef`] reference of the path if the endpoint is present in the interface.
-    pub(crate) fn as_mapping_ref(&self, path: &MappingPath) -> Option<MappingRef<&Interface>> {
+    pub(crate) fn as_mapping_ref<'a>(
+        &'a self,
+        path: &'a MappingPath,
+    ) -> Option<MappingRef<&Interface>> {
```
rgallor, line 189: Maybe I'm missing something, but shouldn't we specify also a lifetime in the return value?

joshuachp: No

rgallor (changes requested):

src/aggregate.rs
```diff
@@ -110,4 +112,12 @@ impl Aggregation {
             None
         }
     }
+
+    /// Returns `true` if the aggregation is [`Unset`].
+    ///
+    /// [`Unset`]: Aggregation::Unset
+    #[must_use]
+    pub fn is_unset(&self) -> bool {
```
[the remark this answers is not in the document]
joshuachp: It's pubblic

src/types.rs
```diff
@@ -515,9 +499,10 @@ impl TryFrom<BsonConverter> for AstarteType {
 
 #[cfg(test)]
 mod test {
-    use crate::Aggregation;
     use chrono::{DateTime, TimeZone, Utc};
 
+    use crate::Aggregation;
```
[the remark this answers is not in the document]
joshuachp: These tests are about the conversions of the `AstarteTypes`.

I've added a test for the functions in aggregation just to increase the coverage

rgallor (approved):

src/aggregate.rs
```diff
@@ -73,6 +73,8 @@ pub enum Aggregation {
     Individual(AstarteType),
     /// Object data, also called aggregate. Can only be from a datastream.
     Object(HashMap<String, AstarteType>),
+    /// Unset of a property
+    Unset,
```
harlem88, line 77: Adding `Unset` under the `Aggregation enum` is a bit strange, I think  we can rename it in `AstartePayload` or 
move Aggregation in another enum that contains `Unset` and `Aggregation`, so :
1.
``` rust
 enum AstartePayload{
   Individual(AstarteType),
   Object(HashMap<String, AstarteType>),
   Unset,
 }
``` 
 2. 
 ``` rust
enum Aggregation{
   Individual(AstarteType),
   Object(HashMap<String, AstarteType>),
 }
 
 enum AstartePaylod {
  Aggregation, 
  Unset,
 }
 ```

harlem88 (changes requested):

src/aggregate.rs
joshuachp: I think the first option whould be more ergonomic to use

joshuachp:
I propose the `Value` for the name of the `Aggregate`, this is the same name in the MQTT protocol (the `v` field of the bson document).

e2e-test/src/main.rs
```diff
@@ -188,7 +189,7 @@ async fn main() {
         match event {
             Ok(data) => {
                 if data.interface == test_cfg.interface_datastream_so {
-                    if let astarte_device_sdk::Aggregation::Individual(var) = data.data {
+                    if let Value::Individual(var) = data.data {
```
rgallor, line 192: you can reduce indentation by using `unwrap_or_else`

rgallor (changes requested):

src/lib.rs
```diff
@@ -182,36 +179,49 @@ impl<S, C> AstarteDeviceSdk<S, C> {
...
             })?;
 
-        let (data, timestamp) = self.connection.deserialize_individual(mapping, payload)?;
+        let individual = self.connection.deserialize_individual(&mapping, payload)?;
 
-        if interface.is_property() {
-            let interface_name = interface.interface_name();
-            let path = path.as_str();
-            let interface_major = interface.version_major();
+        match individual {
```
rgallor, line 184: you can change this into `let...else {}`

rgallor (changes requested):

src/lib.rs
joshuachp: There are cases for both match arms, returning different values, here I like the match statment the most

rgallor (approved):

rgallor (changes requested):

harlem88 (approved):

rgallor (approved):

harlem88 (approved):

PR #297, opened 2024-03-20
fix(grpc): handle correctly recv unset

joshuachp:
Change the `GrpcReceivePayload` data to `Payload` to handle the reception of the `AstarteDataType::Unset` in the individual deserialization.


rgallor (approved):

src/transport/grpc/mod.rs
```diff
@@ -243,18 +244,27 @@ impl Receive for Grpc {
...
         payload: Self::Payload,
     ) -> Result<(AstarteType, Option<Timestamp>), crate::Error> {
-        let astarte_data_type::Data::AstarteIndividual(individual_data) = payload.data else {
-            return Err(crate::Error::from(
-                GrpcTransportError::DeserializationExpectedIndividual,
-            ));
+        let data = match payload.data {
+            Payload::AstarteData(data) => data,
+            Payload::AstarteUnset(astarte_message_hub_proto::AstarteUnset {}) => {
+                trace!("unset received");
```
harlem88, line 250: Do we want to keep it for debugging purposes?

joshuachp: Yes, is also lower than the usual RUST_LOG=debug that we have in other CIs. So it should be disabled in most cases

harlem88 (approved):

harlem88 (approved):

PR #311, opened 2024-03-28
refactor(client): separate client and connection

joshuachp:
Separate the connection in its own struct which is not clone. This permits us to better manage the connection through a mutable reference instead of locking and reference counting it. Use channels to send messages from the client handle to the connection object.


rgallor (changes requested):

src/connection.rs
```diff
@@ -0,0 +1,445 @@
...
+                data: aggregation,
+            });
+
+        self.tx
+            .send_async(data)
+            .await
+            .map_err(|_| Error::Disconnected)
+    }
+
+    pub(crate) async fn handle_client_msgs(&mut self, msg: ClientMessage) -> Result<(), Error>
```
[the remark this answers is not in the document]
joshuachp: This is an internal function. The name `msg` is recognizable and used enough like `req`, `res`, `recv`, etc...  

I like msg more in this case

src/introspection.rs
```diff
@@ -92,10 +85,13 @@ pub trait DynamicIntrospection {
     where
         I: IntoIterator<Item = Interface> + Send;
 
+    /// Add one or more [`Interface`] to the device introspection, specialized for a [`Vec`].
```
rgallor, line 88: I suggest you to change the doc comment at line 80 and make it similar to the ones at line 83 and 89. It is not an error, however it may be misleading (as if the 2 methods are doing different things whereas they are both adding 1 or more `Interface`s)

rgallor (changes requested):

src/builder.rs
```diff
@@ -27,52 +27,70 @@ use std::io;
...
 
 impl<S, C> DeviceSdkBuild<S, C> for DeviceBuilder<S, C>
 where
     S: PropertyStore,
     C: Publish + Receive + Register + Send,
 {
-    fn build(self) -> (AstarteDeviceSdk<S, C>, EventReceiver) {
-        let (tx, rx) = mpsc::channel(self.channel_size);
+    fn build(self) -> (DeviceClient<S>, DeviceConnection<S, C>) {
+        let (tx_connection, rx_client) = flume::bounded(self.channel_size);
```
harlem88, line 69: Is it not possible to use `mpsc::channel(self.channel_size);` also here? What kinds of advantages does `flume` introduce?

src/builder.rs
```diff
@@ -27,52 +27,70 @@ use std::io;
...
+        let interfaces = Arc::new(RwLock::new(self.interfaces));
 
-        (device, rx)
+        let client = DeviceClient::new(
+            Arc::clone(&interfaces),
+            rx_client,
+            tx_client,
+            self.store.clone(),
+        );
+        let device = DeviceConnection::new(
```
harlem88, line 80: `device` -> `connection`

harlem88 (changes requested):

src/builder.rs
joshuachp: The `mpsc` (multi producer single consumer) channel can be used only by one `Receiver` since it doesn't provide clone trait for the consumer part. While flume provides a `mpmc` (multi producer multi consumer) channel, which in our case the single producer of the connection can send the `DeviceEvent` from Astarte to all the `Client` though this queue/channel. This let us have a single clonable struct, the `Client`, to both send and receive data from Astarte. 

The only thing to note with flume is that an event can be received only once, so it has to be handled by the receiver. Nevertheless, this is still fine since the users of the SDK (e.g. Edgehog) will use only one Client to receive the `DeviceEvent`, but are still able to clone the client freely.

I think this should be better documented. So, if it's fine, I will add my reasoning to the PR

joshuachp: The [Tokyo broadcast channel](https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html) is also a mpmc channel implementation, but it suffers from the "slow receiver" problem.

rgallor (approved):

harlem88 (approved):

harlem88 (approved):

PR #313, opened 2024-03-29
fix(store): remove interface's properties

### LINKED ISSUE #312: Delete all interface properties after it's removed
The removal of all the properties of an interface uses the wrong mapping and it will not delete the properties.

This functionality should be implemented in the store, to be handled correctly.

joshuachp:
Implement in the store a way to remove all the properties for an interface. This fixes the deletion of the properties when an interface is removed, since before it was using the mapping without the parameter.

Closes #312


rgallor (approved):

harlem88 (commented):
LGTM, Add some lines to the changelog about this fix

rgallor (approved):

harlem88 (approved):

PR #320, opened 2024-04-09
Feature/interface

joshuachp:


rgallor:
LGTM even though clippy beta fails in CI

rgallor (approved):

src/interface/def.rs
```diff
@@ -343,22 +414,39 @@ impl Display for Aggregation {
...
     DoubleArray,
+    /// Integer array mapping.
     IntegerArray,
+    /// Boolean array mapping.
     BooleanArray,
+    /// Long integer array mapping.
     LongIntegerArray,
+    /// String array mapping.
     StringArray,
+    /// String array mapping.
```
harlem88, line 447:
```suggestion
    /// Binary array mapping.
```

harlem88 (changes requested):

harlem88 (approved):

PR #327, opened 2024-04-22
Dynamic introspection

rgallor:
Update the `DynamicIntrospection` trait with the `remove_interfaces` method, reflecting the new protobuf definition for the Message Hub


src/connection.rs
```diff
@@ -306,6 +313,48 @@ impl<S, C> DeviceConnection<S, C> {
...
+        self.connection
+            .remove_interfaces(&interfaces, &to_remove)
+            .await?;
+
+        for (iface_name, _) in to_remove.iter() {
+            // We cannot error here since we already unsubscribed to the interface
+            if let Err(err) = self.store.delete_interface(iface_name).await {
+                error!("failed to remove property {err}");
+            }
+        }
```
joshuachp, line 349: Check if the interface is a property with as_prop or is_prop

src/interfaces.rs
```diff
@@ -91,6 +93,12 @@ impl Interfaces {
         self.interfaces.remove(interface_name)
     }
 
+    /// Remove the interfaces which name is in the HashSet
+    pub(crate) fn remove_many(&mut self, to_remove: &HashSet<&str>) {
+        self.interfaces
+            .retain(|k, _v| !to_remove.contains(k.as_str()));
+    }
```
joshuachp, line 100: This is just to give and idea, we should do something like this for the following reasons:

- do not allocate an additional set
- the to_remove length set will always be less than the interfaces

```suggestion
    pub(crate) fn remove_many<I, S>(&mut self, to_remove: I) where I: IntoIter<Item = S>, S: AsRef<str> {
        for i in to_remove {
           self.remove(i.as_ref());
        }
    }
```

src/connection.rs
```diff
@@ -306,6 +313,48 @@ impl<S, C> DeviceConnection<S, C> {
...
+
+        let removed_names = to_remove.keys().map(|k| k.to_string()).collect_vec();
+
+        interfaces.remove_many(&HashSet::from_iter(interfaces_name));
```
joshuachp, line 353: With the changes on the remove many of interfaces, this can be simplified in `to_remove.keys()` without allocating the hashset

joshuachp (changes requested):

src/connection.rs
rgallor: It's `&HashSet::from_iter(interfaces_name)` that changes, not `removed_names`, but I got your point. Thanks

src/interface/mod.rs
```diff
@@ -29,6 +29,8 @@ pub(crate) mod validation;
 use log::info;
 use serde::{Deserialize, Serialize};
 
+#[cfg(feature = "message-hub")]
```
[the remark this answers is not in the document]
rgallor: could you please be more explicit?

harlem88 (changes requested):

src/connection.rs
```diff
@@ -306,6 +311,52 @@ impl<S, C> DeviceConnection<S, C> {
...
+    async fn remove_interfaces<'a, I>(&mut self, interfaces_name: I) -> Result<Vec<String>, Error>
+    where
+        C: Register,
+        S: PropertyStore,
+        I: IntoIterator<Item = &'a str> + Clone + Send,
+    {
+        let mut interfaces = self.interfaces.write().await;
+
+        let to_remove: HashMap<&str, &Interface> = interfaces_name
+            .clone()
```
joshuachp, line 323: Remove the clone

joshuachp: Also in the trait bound

src/connection.rs
```diff
@@ -306,6 +311,52 @@ impl<S, C> DeviceConnection<S, C> {
...
+            if let Some(prop) = iface.as_prop() {
+                if let Err(err) = self.store.delete_interface(prop.interface_name()).await {
+                    error!("failed to remove property {err}");
+                }
+            }
+        }
+
+        let removed_names = to_remove.keys().map(|k| k.to_string()).collect_vec();
+
+        interfaces.remove_many(interfaces_name);
```
joshuachp, line 355: Here it should probably be more correct to pass the &removed_names

src/interfaces.rs
```diff
@@ -211,14 +222,24 @@ impl Interfaces {
...
     ) -> impl Iterator<Item = &'a Interface> + Clone {
         self.interfaces
             .values()
             .filter(|i| i.interface_name() != removed.interface_name())
     }
+
+    /// Iter with removed interfaces
+    pub(crate) fn iter_without_removed_many<'a>(
+        &'a self,
+        removed: &'a HashMap<&'a str, &'a Interface>,
```
joshuachp, line 233: Are all the lifetimes here necessary, Can you not remove the ones in the hashmap generic?

src/store/mod.rs
```diff
@@ -46,7 +46,7 @@ pub mod wrapper;
 #[async_trait]
 pub trait PropertyStore: Clone + Debug + Send + Sync + 'static
 where
-    // NOTE: the bounds are required to be compatible with the tokio tasks, with an additional Sync
+    // NOTE: the bounds are required to be compatible with the tokio tasks, with an interface Sync
     //       bound to further restrict the error type.
```
joshuachp, line 50: Refactor gone wrong?

joshuachp (changes requested):

src/mock.rs
joshuachp, line 1: Remove the empty file

src/transport/grpc/mod.rs
```diff
@@ -308,44 +293,80 @@ impl Receive for Grpc {
...
         added: &interfaces::Validated,
     ) -> Result<(), crate::Error> {
-        let iter = interfaces.iter_with_added(added);
-        let data = NodeData::try_from_iter(&self.uuid, iter)?;
+        let interface_json = serde_json::to_string(added.deref())
+            .map_err(|err| Error::Grpc(GrpcError::InterfacesSerialization(err)))?;
+
+        let interface_jsons = InterfacesJson {
+            interfaces_json: vec![interface_json],
+        };
```
joshuachp, line 304: I believe the serialization should already be in the astarte-message-hub-proto

src/store/mod.rs
rgallor: Yep, my mistake

src/connection.rs
```diff
@@ -306,6 +306,51 @@ impl<S, C> DeviceConnection<S, C> {
...
+        if to_remove.is_empty() {
+            return Ok(Vec::new());
+        }
+
+        self.connection
+            .remove_interfaces(&interfaces, &to_remove)
+            .await?;
+
+        for (_, iface) in to_remove.iter() {
+            // We cannot error here since we already unsubscribed to the interface
```
joshuachp, line 339: This comment is referring to the inner if, also I wrote it in an awful English 🙇🏻. If you can reword  this and the above one

```suggestion
            // We cannot error here since we have already unsubscribed from the interface
```

joshuachp (changes requested):

src/transport/mqtt/mod.rs
```diff
@@ -615,6 +609,43 @@ impl Register for Mqtt {
...
+        let introspection = Introspection::new(interfaces).to_string();
+        self.send_introspection(introspection).await?;
+
+        let interfaces_name = removed
+            .values()
+            .filter(|iface| iface.ownership().is_server())
+            .map(|iface| iface.interface_name())
+            .collect_vec();
+
+        self.unsubscribe_many(interfaces_name).await?;
```
joshuachp, line 645: You can just iterate here with a filter or if and not collect into a vector

joshuachp (changes requested):

joshuachp (approved):

harlem88 (approved):

Cargo.toml
```diff
@@ -106,7 +106,7 @@ interface-strict = []
 [workspace.dependencies]
 astarte-device-sdk = { path = "./", version = "=0.8.1" }
 astarte-device-sdk-derive = { version = "=0.8.1", path = "./astarte-device-sdk-derive" }
-astarte-message-hub-proto = "0.6.2"
+astarte-message-hub-proto = { git = "https://github.com/rgallor/astarte-message-hub-proto", branch = "dynamic-introspection" }
```
harlem88, line 109: Now you can change to master branch of official repo

harlem88 (changes requested):

harlem88 (approved):

PR #334, opened 2024-04-30
fix(interfaces): correct the iteration logic

joshuachp:
There was a bug in the iteration that would iterate only on the single added interface in the method Interfaces::iter_with_added. This will generate an incorrect introspection for MQTT connection.


rgallor (approved):

harlem88 (approved):
LGTM. Add fix to the Changelog

harlem88 (approved):

PR #340, opened 2024-05-17
Refactor detach

rgallor:
Use Empty type rather than Node to detach a Node

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Update the Dynamic Introspection to support adding or removing interfaces from a MessageHub Node [#330](https://github.com/astarte-platform/astarte-device-sdk-rust/issues/330)
 
+### Changed
+- Use Empty type rather than Node to detach a Node [#340](https://github.com/astarte-platform/astarte-device-sdk-rust/pull/340/).
+
 ## [0.8.2] - 2024-05-29
 
 ## [0.7.4] - 2024-05-27
--- a/src/transport/grpc/mod.rs
+++ b/src/transport/grpc/mod.rs
@@ -261,12 +261,14 @@ impl Grpc {
             .map_err(GrpcError::from)
     }
 
-    async fn detach(mut client: MsgHubClient, uuid: &Uuid) -> Result<(), GrpcError> {
+    async fn detach(mut client: MsgHubClient) -> Result<(), GrpcError> {
         // During the detach phase only the uuid is needed we can pass an empty array
         // as the interface_json since the interfaces are already known to the message hub
         // this api will change in the future
         client
-            .detach(Node::new(uuid, Vec::new()))
+            .detach(tonic::Request::new(
+                astarte_message_hub_proto::pbjson_types::Empty {},
+            ))
             .await
             .map(|_| ())
             .map_err(GrpcError::from)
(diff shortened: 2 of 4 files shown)


src/transport/grpc/mod.rs
```diff
@@ -965,7 +963,7 @@ mod test {
                 if data_event.interface == "org.astarte-platform.rust.examples.individual-properties.DeviceProperties"
                 && data_event.path == "/1/name"
                 && matches!(data_event.data, Value::Individual(AstarteType::String(v)) if v == STRING_VALUE),
-            ServerReceivedRequest::Detach(d) if d.uuid == ID.to_string()
+            ServerReceivedRequest::Detach(a) if a == astarte_message_hub_proto::pbjson_types::Empty{}
```
joshuachp, line 966: Remove this if, or move it to the match. It's already ensured by the type signature

joshuachp: Same in the other locations

joshuachp (changes requested):

src/transport/grpc/mod.rs
```diff
@@ -906,7 +904,7 @@ mod test {
             => ordered = {i.names.sort(); i.names} ;
                 if ordered == expect_removed,
             // detach
-            ServerReceivedRequest::Detach(a) if a.uuid == ID.to_string()
+            ServerReceivedRequest::Detach(a) if a == astarte_message_hub_proto::pbjson_types::Empty{}
```
joshuachp, line 907:
```suggestion
            ServerReceivedRequest::Detach(Empty {})
```

src/transport/grpc/mod.rs
```diff
@@ -478,7 +477,7 @@ mod test {
     enum ServerReceivedRequest {
         Attach(Node),
         Send(AstarteMessage),
-        Detach(Node),
+        Detach(astarte_message_hub_proto::pbjson_types::Empty),
```
joshuachp, line 480: I would just import the Empty struct since it's used in lot's of places.

joshuachp (changes requested):

src/transport/grpc/mod.rs
rgallor: Yep, could be a good idea

harlem88 (commented):
Add `detach` change to the Changelog

harlem88 (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Update the Dynamic Introspection to support adding or removing interfaces from a MessageHub Node [#330](https://github.com/astarte-platform/astarte-device-sdk-rust/issues/330)
 
+### Changed
+- Use Empty type rather than Node to detach a Node.
```
joshuachp, line 12:
```suggestion
- Use Empty type rather than Node to detach a Node. [#340](https://github.com/astarte-platform/astarte-device-sdk-rust/pull/340/)
```

I think a link to the PR can be really helpful

src/transport/grpc/mod.rs
```diff
@@ -398,9 +398,9 @@ impl Reconnect for Grpc {
 #[async_trait]
 impl Disconnect for Grpc {
     async fn disconnect(mut self) -> Result<(), crate::Error> {
-        Self::detach(self.client, &self.uuid)
-            .await
-            .map_err(|e| e.into())
+        debug!("detaching node {}", self.uuid);
```
joshuachp, line 401:
```suggestion
        info!("detaching node {}", self.uuid);
```

joshuachp: Maybe a debug before and an info after if the detach was successful could be helpful while debugging vs on the debug where the RUST_LOG would be higher than debug.

joshuachp (changes requested):
Since we are updating a git dependency, [pin the rev in the Cargo.toml](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#choice-of-commit)

harlem88 (approved):

joshuachp (approved):

sorru94 (approved):

PR #355, opened 2024-05-29
Feature/mqtt connection

joshuachp:


src/transport/mqtt/connection.rs
```diff
@@ -0,0 +1,760 @@
...
+    }
+}
+
+/// This cannot be a type state machine, because any additional data cannot be moved out of the enum
+/// when polling. The only data that can be easily changed is the current state into the next one.
+#[derive(Debug)]
+enum State {
+    Disconnected(Disconnected),
+    Connecting(Connecting),
+    StartInit(StartInit),
```
harlem88, line 196: What do you think to find a better name for `StartInit` state?
- ProtocolHandshake
- AstarteHandshake
- AstarteMQTTInit
- AstarteInit
- ...

harlem88 (changes requested):

src/transport/mqtt/connection.rs
joshuachp: Yes, something like `Handshake` is more explanatory.

src/connection.rs
```diff
@@ -84,145 +84,233 @@ pub trait ClientDisconnect {
...
+        } = self;
 
-        let interfaces = self.interfaces.read().await;
-        let interface = interfaces.get(interface).ok_or_else(|| {
-            warn!("publish on missing interface {interface} ({path})");
-            Error::InterfaceNotFound {
-                name: interface.to_string(),
+        let mut tasks: JoinSet<Result<(), Error>> = JoinSet::new();
+
+        tasks.spawn(async move {
```
harlem88, line 142: What do you think to introduce a name for each task? So in case of debugging an error we can distinguish which of task generate the error, or is It already possible with the log that we have?

```suggestion
        task::Builder::new()
            .name(client-receiver-task")
            .spawn(async move {
```

src/transport/grpc/mod.rs
```diff
@@ -90,9 +91,9 @@ pub enum GrpcError {
     MessageHubProtoConversion(#[from] MessageHubProtoError),
 }
 
-type MessageHubClientWithInterceptor =
-    MessageHubClient<InterceptedService<Channel, NodeIdInterceptor>>;
+type MhClient = MessageHubClient<InterceptedService<Channel, NodeIdInterceptor>>;
```
harlem88, line 94: Is too short, better use `MsgHubClient`
```suggestion
type MsgHubClient = MessageHubClient<InterceptedService<Channel, NodeIdInterceptor>>;
```

harlem88 (changes requested):

src/connection.rs
joshuachp: We can use a function instead of the async closure. So it will appear in the log traces

rgallor (commented):
It seems good!

rgallor (changes requested):

src/connection.rs
```diff
@@ -84,145 +84,233 @@ pub trait ClientDisconnect {
...
+            // TODO: consider adding a cancellation token.
```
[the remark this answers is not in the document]
joshuachp: Yes, we cannot disconnect for the time being, or stop the two tasks.

src/transport/mqtt/connection.rs
```diff
@@ -0,0 +1,892 @@
...
+
+            if let Some(publish) = opt_publish {
+                self.buff.push_back(publish);
+            }
+
+            // Check if an error occurred
+            if self.state.is_disconnected() {
+                let timeout = exp_back.next();
+
+                debug!("waiting {timeout} seconds before retring");
```
rgallor, line 214: typo "retrying"

src/transport/mqtt/connection.rs
```diff
@@ -0,0 +1,892 @@
...
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) struct Disconnected;
+
+impl Disconnected {
+    /// Recreate the MQTT transport (TLS) to the broker.
+    ///
+    /// It recreate the credentials and reconnect to the broker, using the same
```
rgallor, line 340: typo "recreates"

src/transport/mqtt/connection.rs
```diff
@@ -0,0 +1,892 @@
...
+        let State::Connected(connected) = &mut self.state else {
+            return None;
+        };
+
+        loop {
+            match connected.poll(&mut self.connection).await {
+                Next::Same => {}
+                Next::Publish(publish) => return Some(publish),
+                Next::State(next) => {
+                    debug_assert!(!next.is_connected());
```
rgallor, line 152: is it used for a test/mock expectation?

rgallor (changes requested):

src/transport/mqtt/connection.rs
joshuachp: No, it's just an assert. 

If you read the line above, there is a let else for the state that must be connected. Here, we make sure that we never return to a next state that is the same as the current, since we should have returned a Next::Same. 

Furthermore, if you look below, this match branch would return from the loop, but we must return only when disconnected or in other words a next state different from connected.

rgallor: Thank you for the explanation, now it's clearer :)

src/builder.rs
```diff
@@ -91,7 +91,10 @@ pub enum BuilderError {
 ///
 /// This trait is already implemented generically for the [`DeviceBuilder`]
 /// and implementing it should be avoided since it has no practical use.
-pub trait DeviceSdkBuild<S, C> {
+pub trait DeviceSdkBuild<S, C>
```
harlem88, line 94: Is it possible to  move `S and C` as Associated types to improve the overall readability ?

```suggestion
pub trait DeviceSdkBuild {
    type S;
    type C;
```

joshuachp: I'm unsure if this approach will work, as it would require the builder to implement this trait for a specific associated type, rather than being generic. The implementation could be for some generic parameter, but at that point, the code would essentially be the same as having generics in the trait, since we cannot remove them from the `DeviceBuilder` struct.

Here's an example:

<details title="example">

```rust

pub trait DeviceSdkBuild {
  type Store;
  type Conn;
}

impl<S,C> DeviceSdkBuilder for DeviceBuilder<S,C> {
  type Store = S;
  type Conn = C;
// ...
}

```

</details>

Also, we use generics to avoid exposing the underlying private types. However, implementing this trait would still expose those types.

src/connection.rs
```diff
@@ -85,145 +85,234 @@ pub trait ClientDisconnect {
...
+                };
 
-        match individual {
-            Some((value, timestamp)) => {
-                if let Some(prop) = mapping.as_prop() {
-                    let prop = StoredProp::from_mapping(&prop, &value);
+                self.store.store_prop(prop).await?;
 
-                    self.store.store_prop(prop).await?;
+                info!(
```
harlem88, line 239:
```suggestion
                debug!(
```

src/connection.rs
```diff
@@ -351,166 +436,144 @@ impl<S, C> DeviceConnection<S, C> {
...
+            None => {
+                // Unset can only be received for a property
+                self.store
+                    .delete_prop(interface.interface_name(), path.as_str())
+                    .await?;
 
-                if let Err(Err(err)) = response.send(res) {
-                    error!(error = %Report::new(err), "client disconnected while failing to remove interfaces");
-                }
+                info!(
```
harlem88, line 527:
```suggestion
                debug!(
```

src/transport/mqtt/connection.rs
```diff
@@ -0,0 +1,893 @@
...
+
+            if let Some(publish) = opt_publish {
+                self.buff.push_back(publish);
+            }
+
+            // Check if an error occurred
+            if self.state.is_disconnected() {
+                let timeout = exp_back.next();
+
+                debug!("waiting {timeout} seconds before retying");
```
harlem88, line 215:
```suggestion
                debug!("waiting {timeout} seconds before retrying");
```

src/transport/mqtt/connection.rs
```diff
@@ -0,0 +1,893 @@
...
+
+        mqtt_connection
+            .reconnect(client_id, interfaces, store)
+            .await?;
+
+        Ok(mqtt_connection)
+    }
+
+    /// Connect to astarte, wait till the state is [`Connected`].
+    pub(crate) async fn reconnect<S>(
```
harlem88, line 188: Why did you call it `reconnect`? It seems more like something about the first connection

src/transport/mqtt/connection.rs
```diff
@@ -0,0 +1,893 @@
...
+    {
+        trace!("state {}", self);
+
+        let next = match self {
+            State::Disconnected(disconnected) => disconnected.reconnect(conn, client_id).await,
+            State::Connecting(connecting) => connecting.wait_connack(conn).await,
+            State::Handshake(start_init) => {
+                let session_data = SessionData::try_from_props(interfaces, store).await?;
+
+                start_init.init(conn, client_id, session_data)
```
harlem88, line 285:
```suggestion
            State::Handshake(handshake) => {
                let session_data = SessionData::try_from_props(interfaces, store).await?;

                handshake.init(conn, client_id, session_data)
```

harlem88 (changes requested):

src/transport/mqtt/connection.rs
joshuachp: Because we don't start in a disconnected state, so we have being disconnected and we are reconnecting. The first state is the `Connecting` state.

joshuachp: Yeh, in an earlier version, the wait_connack and reconnect were two different functions now it makes more sens to call it `connect`

harlem88 (approved):

rgallor (approved):

PR #362, opened 2024-07-18
Reflect the message hub proto api changes

rgallor:
Since the Message Hub server can now send custom errors through the gRPC connection (and not only AstarteMessages), the sdk changes are meant to provide this error information to the client.


src/client.rs
```diff
@@ -43,6 +45,50 @@ use crate::{
...
+    },
+
+    /// Couldn't find missing mapping in the interface.
+    #[error("couldn't find mapping {mapping} in interface {interface}")]
+    MappingNotFound {
+        /// Name of the interface.
+        interface: String,
+        /// Path of the missing mapping.
+        mapping: String,
+    },
```
harlem88, line 74: Is `MappingNotFound` an error that can come also from `MapppingError`?

src/client.rs
```diff
@@ -43,6 +45,50 @@ use crate::{
...
+
+    /// Invalid aggregation between the interface and the data.
+    #[error("invalid aggregation, expected {exp} but got {got}")]
+    Aggregation {
```
harlem88, line 78: Does it represent all interface types, or only the Datastream (aggregation individual/object) type?

harlem88 (changes requested):

src/client.rs
rgallor: No, it is the equivalent of the `MappingNotFounf` variant you can find in the `src/error.rs` file

src/client.rs
rgallor: It represents only the Individual/Object datastream type.

harlem88: Since sometimes the properties are handled as Individual type, If there is an error on `property` side will an Aggregation error be used, or should we add an specific error?

rgallor: I added some context to the `Aggregate` error variant, so now the distinction should be clearer

joshuachp (changes requested):

src/connection.rs
```diff
@@ -563,13 +602,22 @@ impl<S, C> DeviceReceiver<S, C> {
...
 
+        // let data = match res {
+        //     Ok(data) => Ok(data),
+        //     Err(err) => {
+        //         // this way, if there is an unrecoverable error the application will return Err
+        //         let recv_err = RecvError::try_from(err)?;
+        //         Err(recv_err)
+        //     }
+        // };
+
```
harlem88, line 620: Remove it. if it is no longer used
```suggestion
```

src/transport/grpc/mod.rs
```diff
@@ -285,20 +285,47 @@ impl std::fmt::Debug for Grpc {
...
+                let Some(res) = message_res.event else {
+                    error!("empty MessageHubEvent should never occur");
+                    return Ok(Some(Err(RecvError::Connection(Box::new(
+                        MessageHubProtoError::ExpectedField("event"),
+                    )))));
+                    // return Ok(ConnectionEvent::error(RecvError::Connection(
+                    //     ConnectionRecvError::ProtoConversion(MessageHubProtoError::ExpectedField(
+                    //         "event",
+                    //     )),
+                    // )));
```
harlem88, line 304:
```suggestion
```

src/transport/grpc/mod.rs
```diff
@@ -285,20 +285,47 @@ impl std::fmt::Debug for Grpc {
...
+                };
+
+                let message = match res {
+                    astarte_message_hub_proto::message_hub_event::Event::Message(msg) => msg,
+                    astarte_message_hub_proto::message_hub_event::Event::Error(proto_err) => {
+                        error!("message hub proto error: {proto_err}");
+                        return Ok(Some(Err(RecvError::Connection(Box::new(proto_err)))));
+                        // return Ok(ConnectionEvent::error(RecvError::Connection(
+                        //     ConnectionRecvError::Server(proto_err),
+                        // )));
```
harlem88, line 314:
```suggestion
```

src/transport/grpc/mod.rs
```diff
@@ -285,20 +285,47 @@ impl std::fmt::Debug for Grpc {
...
+                    .map_err(|err| RecvError::Connection(Box::new(err)));
                 Ok(Some(event))
+
+                // let con_event: ReceivedEvent<Self::Payload> = message
+                //     .try_into()
+                //     .map_err(|err| RecvError::Connection(ConnectionRecvError::ProtoConversion(err)))
+                //     .map(ConnectionEvent::Event)
+                //     .unwrap_or_else(ConnectionEvent::ReceiveError);
+
+                // Ok(con_event)
```
harlem88, line 328:
```suggestion
```

harlem88 (changes requested):
Remove commented code

rgallor:
In this PR I can't update the `astarte-message-hub-proto` version to v0.7.0 since it also contains modification to the Attach rpc, made in the `related PR #372 .

src/client.rs
```diff
@@ -43,6 +45,47 @@ use crate::{
     AstarteAggregate, Error, Interface,
 };
 
+type DynError = Box<dyn std::error::Error + Send + Sync + 'static>;
```
joshuachp, line 48: The `DynError` is also used in the store  (and maybe other places), we can move it to the src/error.rs file and make it pub(crate)

src/client.rs
```diff
@@ -43,6 +45,47 @@ use crate::{
...
+
+/// Possible errors returned while handling connection messages
+#[non_exhaustive]
+#[derive(thiserror::Error, Debug)]
+pub enum RecvError {
+    /// Connection error, either gRPC or MQTT
```
joshuachp, line 54:
```suggestion
    /// A generic connection related error. 
    /// 
    /// Should be downcasted to access the underling specific connection error.
```

src/transport/mod.rs
```diff
@@ -82,6 +83,12 @@ pub(crate) trait Publish {
     async fn unset(&mut self, data: ValidatedUnset) -> Result<(), crate::Error>;
 }
 
+pub(crate) type EventResult<P> = Result<Option<Result<ReceivedEvent<P>, RecvError>>, crate::Error>;
+pub(crate) type DeserializeIndResult =
+    Result<Result<Option<(AstarteType, Option<Timestamp>)>, RecvError>, crate::Error>;
+pub(crate) type DeserializeObjResult =
```
joshuachp, line 89: This is really confusing. We are having multiple layers of errors and values.

This signature can be flattened by grouping the errors. The result should be:

```rust
pub(crate) enum TransportError {
    /// Error that will be sent to the client
    Recv(#[from] RecvError),
    /// Error from the underline transport
    Transport(#[source] crate::Error),
}

fn foo() -> Result<Option<T>, ConnectionError> {...}
```

joshuachp (changes requested):

harlem88 (approved):

joshuachp (approved):

harlem88 (approved):

PR #363, opened 2024-07-18
feat(retention): implement the retention for the connestions

joshuachp:
Create the structures needed to store the packets for the interfaces with reliability stored.


joshuachp:
I will separate the PR in smaller commits, and adjustments for the dynamic introspection are still required.

src/retention/sqlite.rs
```diff
@@ -0,0 +1,973 @@
...
+            }
+
+            warn!("mappings differ, replacing");
+        }
+
+        let retention: u8 = match mapping.reliability {
+            Reliability::Unreliable => 0,
+            Reliability::Guaranteed => 1,
+            Reliability::Unique => 2,
+        };
```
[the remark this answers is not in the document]
joshuachp: This is not guarantied without making the enum repr(u8) or giving explicit values

src/retention/mod.rs
```diff
@@ -0,0 +1,425 @@
...
+            &individual.interface,
+            &individual.path,
+            individual.version_major,
+            individual.reliability,
+            individual.retention,
+            value,
+        )
+    }
+
+    fn from_obj(obj: &'a ValidatedObject, value: &'a [u8]) -> Self {
```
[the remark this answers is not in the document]
joshuachp: I would prefer obj since is an internal function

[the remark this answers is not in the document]
joshuachp: I actually would like to find a proper abbreviation for individual, but alas no one came to mind

migrations/20240513131630_retention.sql
```diff
@@ -0,0 +1,40 @@
...
+    -- Quality of service
+    reliability INTEGER NOT NULL,
+    -- Seconds after the entry will expire
+    expiry_sec INTEGER,
+    PRIMARY KEY (interface, path)
+);
+
+
+-- Interface stored retention
+CREATE TABLE IF NOT EXISTS retention_publish (
```
harlem88, line 18: These two table have the same description

src/connection.rs
```diff
@@ -309,6 +362,114 @@ impl<S, T> DeviceSender<S, T> {
...
+    where
+        T: Publish,
+    {
+        if !self.status.is_connected() {
+            trace!("publish individual while connection is offline");
+
+            return self.offline_send_individual(data).await;
+        }
+
+        if data.retention.is_stored() {
```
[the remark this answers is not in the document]
joshuachp: This is a good point. For both volatile and stored retentions, I don't remove the inflight expired messages.

The MQTT library automatically Acks the packages, so for now I skip the expiration of inflight packages.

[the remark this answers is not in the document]
joshuachp: The volatile now behaves like the stored and better handles the expiration of packets.

src/transport/mqtt/mod.rs
```diff
@@ -563,13 +613,8 @@ where
     {
         static PURGE_PROPERTIES_TOPIC: OnceCell<String> = OnceCell::new();
 
-        self.mark_received_packets().await?;
-
-        // We need to construct the ClientId ourselves to not borrow self wail calling the method.
-        let client_id = self.client_id.as_ref();
-
         // Wait for next data or until its disconnected
```
rgallor, line 664: typo `its`

rgallor (changes requested):

src/transport/mqtt/config/mod.rs
```diff
@@ -365,21 +365,26 @@ impl MqttConfig {
 
         mqtt_opts.set_transport(transport);
 
+        // Set the clean_session since this is the first connection.
+        mqtt_opts.set_clean_session(true);
```
rgallor, line 369: Also when the reconnect() is called, clean session will be set to true (since). Is this the behavior we want?

harlem88 (changes requested):

src/retention/memory.rs
```diff
@@ -0,0 +1,554 @@
...
+}
+
+impl Default for VolatileStore {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+struct VolitileItem {
```
harlem88, line 195: Why didn't you reuse `PublishInfo` struct from the module for the items stored in volatile memory? Maybe some lines of code can be eliminated.

src/transport/mqtt/config/mod.rs
joshuachp: Now we reset the clean_session upon successful connection or publish

src/retention/memory.rs
joshuachp: I don't immediately see where the PublishInfo could be used for the volatile store.

The VolatileItem has additional information for the store time and sent flag, which are missing from the PublishInfo. Furthermore, the PublishInfo requires the payload to be serialized befeore storing it, while for the VolatileItem we store the AstarteType or Object.

harlem88 (changes requested):

harlem88 (commented):
LTGM. Add the retention feature to the Changelog.

joshuachp:
Changelog updated

harlem88 (approved):

src/retention/memory.rs
```diff
@@ -0,0 +1,554 @@
...
+        };
+
+        (self.store_time + expiry) < now
+    }
+}
+
+/// Failed to store publish information for interface without volatile retention.
+#[non_exhaustive]
+#[derive(thiserror::Error, Debug)]
+#[error("interface {interface} has't retention volatile, but has {retention:?} instead")]
```
rgallor, line 224: typo has't -> `doesn't have`

rgallor (commented):
LGTM

rgallor (approved):

rgallor (approved):

harlem88 (approved):

PR #364, opened 2024-07-30
chore(deps): pin message hub proto version for dynamic introspection

rgallor:
This is necessary to avoid deps conflicts in the Message Hub


joshuachp (approved):
📌

harlem88 (approved):

PR #367, opened 2024-08-09
feat(sqlite): move to the rusqlite crate

joshuachp:


rgallor (changes requested):

src/retention/sqlite/statements.rs
```diff
@@ -0,0 +1,1103 @@
...
+    ) -> Result<(), SqliteError> {
+        let store_mapping =
+            read_mapping(self, &mapping.interface, &mapping.path)?.is_some_and(|stored| {
+                if stored != *mapping {
+                    warn!("mappings differ, replacing");
+
+                    true
+                } else {
+                    false
+                }
```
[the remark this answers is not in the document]
joshuachp: The log is fundamental, the mapping must never differ and thus the warning if it happens

src/store/sqlite/mod.rs
```diff
@@ -240,82 +247,134 @@ impl TryFrom<StoredRecord> for StoredProp {
...
+    ///     let store = SqliteStore::connect("/val/lib/astarte/").await.unwrap();
     /// }
     /// ```
-    pub async fn from_uri(uri: &str) -> Result<Self, SqliteError> {
-        let options = SqliteConnectOptions::from_str(uri).map_err(|err| SqliteError::Uri {
-            backtrace: err,
-            uri: uri.to_string(),
-        })?;
+    pub async fn connect(writable_path: impl AsRef<Path>) -> Result<Self, SqliteError> {
+        // TODO: rename the database to store.db since it doesn't contain only  properties
```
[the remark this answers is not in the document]
joshuachp: No, it's no inherent to the PR and not straight forward.

src/retention/sqlite/statements.rs
```diff
@@ -0,0 +1,1124 @@
...
+            .map_err(SqliteError::Query)?;
+
+        Ok(())
+    }
+
+    pub(super) fn store_publish(
+        transaction: &Transaction<'_>,
+        publish: &RetentionPublish<'_>,
+    ) -> Result<(), SqliteError> {
+        let mut statement = transaction
```
harlem88, line 95: Why didn't you wrap this part with `wrap_sync_call` as well?

joshuachp: Because the store_publish and store_mapping are lower "level" functions for the `store` call, and we can wrap both in a single wrap_sync_call.

This should be private, but I exposed it to be used in the tests.

src/store/sqlite/statements.rs
```diff
@@ -0,0 +1,311 @@
...
+    }
+}
+
+impl Deref for ReadConnection {
+    type Target = Connection;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
```
harlem88, line 311: What do you think about adding some test to this module?

joshuachp: This file is fairly covered by our "store tests" as you can see:

- Here is the invocation https://github.com/joshuachp/astarte-device-sdk-rust/blob/cfc15ad535a20ae60ff6e56085af5a24a88048be/src/store/sqlite/mod.rs#L540
- And the tests are here https://github.com/joshuachp/astarte-device-sdk-rust/blob/a57661a93acd8d7f4485e72ffaf8bf368f26daf5/src/store/mod.rs#L181

The resulting coverage is pretty high (nearly 90%) https://app.codecov.io/gh/astarte-platform/astarte-device-sdk-rust/pull/367/blob/src/store/sqlite/statements.rs#src/store/sqlite/statements.rs.

Most of the missing lines are for query errors. Those are harder to test since the query are correct.

Should I port some of the test from the store to this module just in case?

harlem88: If it is covered by `store tests` it is ok, we no need to add any other tests.

rgallor (approved):

harlem88 (approved):

PR #372, opened 2024-08-27
Use node id metedata in the attach rpc

rgallor:
Reflect the astarte-message-hub-proto changes: the node id information has been removed from the attach rpc parameter, which now contains only the introspection.

### DIFF
--- a/src/transport/grpc/mod.rs
+++ b/src/transport/grpc/mod.rs
@@ -481,7 +481,7 @@ where
 {
     async fn reconnect(&mut self, interfaces: &Interfaces) -> Result<(), crate::Error> {
         // try reattaching
-        let data = NodeData::try_from_interfaces(self.uuid, interfaces)?;
+        let data = NodeData::try_from(interfaces)?;
 
         let mut exp_back = ExponentialIter::default();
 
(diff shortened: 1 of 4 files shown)


src/transport/grpc/mod.rs
```diff
@@ -610,17 +610,17 @@ struct NodeData {
...
+        let node = Node::from_interfaces(interfaces)?;
 
         Ok(Self { node })
     }
 
-    fn try_from_interfaces(uuid: Uuid, interfaces: &Interfaces) -> Result<Self, GrpcError> {
-        Self::try_from_iter(uuid, interfaces.iter())
+    fn try_from_interfaces(interfaces: &Interfaces) -> Result<Self, GrpcError> {
+        Self::try_from_iter(interfaces.iter())
     }
```
joshuachp, line 624: Implement TryFrom here.

src/transport/grpc/mod.rs
```diff
@@ -649,9 +649,11 @@ mod test {
 
     use super::*;
 
+    const ID: Uuid = uuid!("67e55044-10b1-426f-9247-bb680e5fe0c8");
+
     #[derive(Debug)]
     enum ServerReceivedRequest {
-        Attach(Node),
+        Attach((Uuid, Node)),
```
harlem88, line 656: Why did you add uuid here? Is not possible to read from `get_node_id_from_metadata`?

joshuachp (changes requested):

harlem88 (approved):

joshuachp (approved):

PR #375, opened 2024-09-09
feat(derive): implement FromEvent for individual

joshuachp:
Derive the FromEvent trait for interfaces with aggregation individual. This will  require the Rust item to be an `enum` with a variant for each of the mapping's endpoint. To distinguish the aggregation, a name-value pair was added to the `#[from_event]` attribute. All the derive macros have also been updated and changed to use the `attributes` instead of providing a different macro to pars them.

### DIFF
--- a/astarte-device-sdk-derive/CHANGELOG.md
+++ b/astarte-device-sdk-derive/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project
 adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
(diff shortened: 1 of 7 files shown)


rgallor (approved):

astarte-device-sdk-derive/CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project
 adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] - 2024-08-22
```
harlem88, line 8:
```suggestion
## [Unreleased]

harlem88 (changes requested):

harlem88 (approved):

PR #377, opened 2024-09-11
feat(prop): handle properties when offline 

joshuachp:
Do not send properties when we are offline, also unset the property
before sending it. This will let us, if offline, resend the unset when
re-connected and delete the property only if actually sent.

### DIFF
--- a/src/store/mod.rs
+++ b/src/store/mod.rs
@@ -80,6 +80,8 @@ where
         path: &str,
         interface_major: i32,
     ) -> Result<Option<AstarteType>, Self::Err>;
+    /// Unset a property from the database.
+    async fn unset_prop(&self, interface: &str, path: &str) -> Result<(), Self::Err>;
     /// Delete a property from the database.
     async fn delete_prop(&self, interface: &str, path: &str) -> Result<(), Self::Err>;
     /// Removes all saved properties from the database.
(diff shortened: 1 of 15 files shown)


src/store/mod.rs
```diff
@@ -121,6 +125,12 @@ pub struct StoredProp<S = String, V = AstarteType> {
     pub ownership: Ownership,
 }
 
+/// A property that may be unset.
+///
+/// This is returned by getting all the properties (`load_all_props`) that have not been deleted
+/// yet, since they where not sent to Astarte.
+pub type MaybeStoredProp = StoredProp<String, Option<AstarteType>>;
```
harlem88, line 132: I would like to suggest a different name for `MaybeStoredProp`:
- `NullableStoredProp`
- `OptionalStoredProp`
- ...

harlem88: Since it represents a property that could be  `Set or Unset`, I think a name with `Opt` is better.
```suggestion
pub type OptStoredProp = StoredProp<String, Option<AstarteType>>;
```

harlem88 (changes requested):

src/store/mod.rs
harlem88: I vote for `OptStoredProp` I think it represents better than `PropWithUnset`.

harlem88 (approved):

rgallor (approved):

PR #378, opened 2024-09-12
feat(derive): add allow unset for Optional values

joshuachp:
Add the attribute `allow_unset` to make the value of an individual property be wrapped in an `Option` when the value is un `Value::Unset`.

### DIFF
--- a/astarte-device-sdk-derive/src/event.rs
+++ b/astarte-device-sdk-derive/src/event.rs
@@ -30,7 +30,7 @@ use syn::{
 };
 
 use crate::{
-    case::RenameRule, parse_attribute_list, parse_name_value_attrs, parse_str_lit,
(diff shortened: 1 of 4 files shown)


rgallor (approved):

astarte-device-sdk-derive/src/event.rs
```diff
@@ -226,8 +226,50 @@ impl FromEventDerive {
...
+                                return Ok(#name::#variant(None));
+                            },
+                            Value::Object(_) => {
+                                return Err(FromEventError::Object {
+                                    interface: INTERFACE,
+                                    endpoint: event.path,
+                                });
+                            }
+                        };
+
```
harlem88, line 244: Remove the double new line
```suggestion
```

harlem88 (changes requested):

harlem88 (approved):

PR #390, opened 2024-10-29
fix(mqtt): send purge device properties on conn

joshuachp:
Send a purge properties to Astarte on first connection to make sure the connection and properties are in good state.


harlem88 (commented):
LGTM. Add an entry to the changelog about this fix

joshuachp:
Added

harlem88 (approved):

PR #395, opened 2024-11-26
Feature/bump rust 1.78

lucaato:


src/transport/mqtt/mod.rs
```diff
@@ -787,26 +780,24 @@ impl SessionData {
     }
 }
 
-#[async_trait]
 trait AsyncClientExt {
     /// Sends the introspection [`String`].
-    async fn send_introspection(
+    fn send_introspection(
```
harlem88, line 785: Is it not possible to keep sugar version `async`, since it is a private trait?
```suggestion
    async fn send_introspection(
    ...
    ) -> Result<NoticeFuture, ClientError>;
```

src/connection.rs
```diff
@@ -627,11 +625,13 @@ where
...
         if let Some(retention) = self.store.get_retention() {
+            // TODO here stuff goes wrong if possible try to avoid using the boxed future
             let res = retention
                 .delete_interface_many(
-                    to_add
+                    &(to_add
                         .values()
-                        .filter_map(|v| v.is_major_change().then_some(v.interface_name())),
+                        .filter_map(|v| v.is_major_change().then_some(v.interface_name()))
+                        .collect::<Vec<_>>()),
```
rgallor, line 634: you can directly use `collect_vec()` from the (already imported) itertools crate instead of `collect::<Vec<_>>())`

src/connection.rs
```diff
@@ -709,7 +709,9 @@ where
             .await?;
 
         if let Some(retention) = self.store.get_retention() {
-            let res = retention.delete_interface_many(to_remove.keys()).await;
+            let res = retention
+                .delete_interface_many(&to_remove.keys().collect::<Vec<_>>())
```
rgallor, line 713: same here

rgallor (changes requested):

Cargo.toml
```diff
@@ -45,7 +45,7 @@ edition = "2021"
 homepage = "https://astarte.cloud/"
 license = "Apache-2.0"
 repository = "https://github.com/astarte-platform/astarte-device-sdk-rust"
-rust-version = "1.72.0"
+rust-version = "1.78.0"
```
joshuachp, line 48: I would check if we can remove the transitive and pinned dependencies

src/builder.rs
```diff
@@ -376,13 +374,18 @@ impl Default for DeviceBuilder<(), ()> {
     }
 }
 
+/// Structure that stores a successfully established connection
+pub struct EstablishedConnection<C: Connection> {
```
joshuachp, line 378: What do you think about renaming this into Transport?

Cargo.toml
lucaato: Removed the transitive dependencies and removed `once-cell`

src/connection.rs
```diff
@@ -627,11 +625,13 @@ where
         self.sender.extend_interfaces(&interfaces, &to_add).await?;
 
         if let Some(retention) = self.store.get_retention() {
+            // TODO here stuff goes wrong if possible try to avoid using the boxed future
             let res = retention
                 .delete_interface_many(
-                    to_add
+                    &(to_add
```
rgallor, line 631: I think you could avoid wrapping inside brackets: use `&` instead of `&(...)`

rgallor (changes requested):
LGTM

Cargo.toml
```diff
@@ -79,12 +77,9 @@ tracing = { workspace = true }
 url = { workspace = true, features = ["serde"] }
 uuid = { workspace = true, features = ["v4", "v5"] }
 
-# Transitive dependency
-ahash = { workspace = true }
-time = { workspace = true }
-
 [dev-dependencies]
 astarte-device-sdk-derive = { workspace = true }
+async-trait = { workspace = true }
```
harlem88, line 82: Is it still necessary for dev-dependencies?

lucaato: Yes because I implemented the MessageHubServer trait to test the grpc connection, and that one is still using the async_trait crate.

harlem88 (approved):

rgallor:
I'd update also the `astarte-message-hub-proto` version (since the MSRV has been updated) and use the git reference

rgallor (approved):

joshuachp (approved):

PR #399, opened 2024-12-18
Feature/get properties

lucaato:


src/transport/grpc/mod.rs
```diff
@@ -622,13 +640,257 @@ impl<'a> TryFrom<&'a Interfaces> for NodeData {
...
+    type Retention = S::Retention;
+
+    fn get_retention(&self) -> Option<&Self::Retention> {
+        self.store.get_retention()
+    }
+}
+
+/// We implement the PropertyStore to override the behavior when retrieving or storing
+/// device owned properies. We only want to store and load from the store server owned properties.
+/// Device owned properties wont be stored and will be requested to the message hub server.
```
rgallor, line 705: typo won't

src/transport/grpc/mod.rs
```diff
@@ -622,13 +640,257 @@ impl<'a> TryFrom<&'a Interfaces> for NodeData {
...
+
+    async fn store_prop(&self, prop: StoredProp<&str, &AstarteType>) -> Result<(), Self::Err> {
+        if prop.ownership.is_server() {
+            self.store
+                .store_prop(prop)
+                .await
+                .map_err(GrpcStoreError::from_err)
+        } else {
+            Ok(())
+        }
```
rgallor, line 721: you could increase readability with:
```suggestion
// we only store server-owned properties
        if prop.ownership.is_device() { // or use   !prop.ownership.is_server()
            return Ok(());
        }

        self.store
            .store_prop(prop)
            .await
            .map_err(GrpcStoreError::from_err)
```

rgallor: the same stands for all the other methods where the else condition returns Ok(())

src/transport/mod.rs
```diff
@@ -67,17 +68,28 @@ pub(crate) struct ReceivedEvent<P> {
...
+    note = "Before calling `crate::builder::DeviceSdkBuild::build` you should call `DeviceBuilder::connect`"
+)]
 pub trait Connection {
     /// Sender for the connection.
     ///
     /// This reduces the number of generics for connection, since a single client type is associated
     /// with a connection.
     type Sender: Send + Sync;
+    // /// Type of the store wrapper
+    //type Store: PropertyStore;
```
rgallor, line 83: this shouldn't be necessary anymore, should it?

rgallor (changes requested):

src/transport/grpc/mod.rs
```diff
@@ -622,13 +640,267 @@ impl<'a> TryFrom<&'a Interfaces> for NodeData {
...
+                .map_err(GrpcStoreError::from_err)
+        };
+
+        let (server_result, device_result) =
+            futures::future::join(load_server, self.load_device_properties()).await;
+
+        Ok(server_result?
+            .into_iter()
+            .chain(device_result?.into_iter())
+            .collect_vec())
```
rgallor, line 832: Just for greater readability, I'd extract this value into a variable and return Ok(var)

src/transport/grpc/mod.rs
```diff
@@ -622,13 +640,267 @@ impl<'a> TryFrom<&'a Interfaces> for NodeData {
...
+    }
+
+    async fn interface_props<I>(
+        &self,
+        interface: &StoreInterfaceData<I>,
+    ) -> Result<Vec<StoredProp>, Self::Err>
+    where
+        I: AsRef<str> + Send + Sync,
+    {
+        // if to get properties only when connecting
```
rgallor, line 853: I don't get this comment

rgallor (changes requested):

src/store/mod.rs
```diff
@@ -49,6 +50,53 @@ pub trait StoreCapabilities {
...
+
+/// Converts a property ref object reference to the store needed input
+impl<'a> From<&'a PropertyRef<'a>> for StoreInterfaceData<&'a str> {
+    fn from(prop_ref: &'a PropertyRef) -> Self {
+        Self::new(prop_ref.0.interface_name(), prop_ref.0.ownership())
+    }
+}
+
+/// Converts a stored prop reference to the store needed input
+impl<'a, S, T> From<&'a StoredProp<S, T>> for StoreInterfaceData<&'a str>
```
rgallor, line 91: You could use here the same Generic notation used in the definition of `StoredProp` (which is <S, V> rather than <S, T>)

rgallor (changes requested):

src/transport/grpc/mod.rs
```diff
@@ -622,13 +640,266 @@ impl<'a> TryFrom<&'a Interfaces> for NodeData {
...
+                .map_err(GrpcStoreError::from_err)
+        };
+
+        let (server_result, device_result) =
+            futures::future::join(load_server, self.load_device_properties()).await;
+
+        let merged_properties = server_result?
+            .into_iter()
+            .chain(device_result?.into_iter())
+            .collect_vec();
```
rgallor, line 830: You could directly use the `append` method of vec (thus avoid importing `Itertools`:

```suggestion
let mut merged_properties = server_result?;
merged_properties.append(&mut device_result?);
```

rgallor (changes requested):

src/transport/grpc/mod.rs
lucaato: This avoids having a mut variable, itertools is already imported anyways, are you sure anyway about the change ?

rgallor: It would have sense using `chain` with iterators if you don't have to collect at the end. Also, append does not clone but rather move elements from one vec to the other

rgallor (approved):

src/transport/grpc/mod.rs
lucaato: In the end I applied the change

lucaato:
After a bit of irl discussion and a bit of help from @joshuachp we refactored out the `WrapStore` trait, along with some changes to the device builder that should be small enough to fit in this pr.

src/store/mod.rs
```diff
@@ -49,6 +50,53 @@ pub trait StoreCapabilities {
     fn get_retention(&self) -> Option<&Self::Retention>;
 }
 
+/// Data passed to the store that identifies an interface
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct StoreInterfaceData<S = String> {
```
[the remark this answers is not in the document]
lucaato: Maybe InterfaceData or InterfaceInfo, I do not especially like Record as a suffix because it makes it look like a db stored record even if that's not the case.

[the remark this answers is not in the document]
lucaato: Settled on IntefaceInfo.

src/transport/grpc/convert.rs
```diff
@@ -62,6 +64,68 @@ pub enum MessageHubProtoError {
     DateConversion(String),
 }
 
+/// Map a received message hub property to an optional astarte type
+pub fn map_property_to_astarte_type(
```
harlem88, line 68: Is this function need to be `pub`? Maybe is keep only visible to the crate
```suggestion
pub(crate) fn map_property_to_astarte_type(
```

src/transport/grpc/convert.rs
```diff
@@ -62,6 +64,68 @@ pub enum MessageHubProtoError {
...
+            .individual_data
+            .ok_or(MessageHubProtoError::ExpectedField("individual_data"))?,
+        astarte_message_hub_proto::property::Value::AstarteUnset(_) => return Ok(None),
+    };
+
+    Ok(Some(individual.try_into()?))
+}
+
+/// Map a list of properties that need to be set, unset value will result in an error of the conversion
+pub fn map_set_stored_properties(
```
harlem88, line 85: same above

src/transport/grpc/mod.rs
```diff
@@ -622,13 +632,260 @@ impl<'a> TryFrom<&'a Interfaces> for NodeData {
     }
 }
 
+/// Store wrapper designed specifically for the grpc connection
```
harlem88, line 635: Grpc Store?
```suggestion
/// GrpcStore designed specifically for the grpc connection
```

harlem88 (changes requested):

rgallor (approved):

src/store/mod.rs
```diff
@@ -74,24 +122,30 @@ where
     ///
     /// The property store should delete the property from the database if the major version of the
     /// interface does not match the one provided.
-    fn load_prop(
+    fn load_prop<I>(
```
rgallor, line 125: you could use generic `I` also in the `PropertyStore` implementation for `MemoryStore` and `SqliteStore` (where instead you are using `S`)

rgallor (changes requested):

src/transport/grpc/mod.rs
lucaato: Since it would have been a bit repetitive, I changed it this way
```suggestion
/// Store implementation designed specifically for the grpc connection
```

src/transport/grpc/convert.rs
lucaato: Yes they can probably be carte public if the message hub does not use them, and I don't think it does.

src/store/mod.rs
lucaato: I think `S` is more correct since the `StoredProp` also uses that notation.

Cargo.toml
```diff
@@ -107,7 +107,7 @@ rustc-args = ["--cfg=docsrs"]
 [workspace.dependencies]
 astarte-device-sdk = { path = "./", version = "=0.9.3" }
 astarte-device-sdk-derive = { version = "=0.9.3", path = "./astarte-device-sdk-derive" }
-astarte-message-hub-proto = { git = "https://github.com/astarte-platform/astarte-message-hub-proto.git", rev = "5ff0672d8fdc8cba5a2ff023895ae505d75c3d1a" }
+astarte-message-hub-proto = { git = "https://github.com/astarte-platform/astarte-message-hub-proto.git", rev = "370be08ff90236a25b15af6e27d5363739566f2f" }
```
rgallor, line 110: You could update the rev commit since the master branch of the astarte-message-hub-proto repo has been updated, however the new modifications do not impact this functionality (so feel free to leave this rev)

rgallor (changes requested):

Cargo.toml
lucaato: I think I will leave it as is and bump it in the test refactor pr

rgallor (approved):

harlem88 (approved):

src/interfaces.rs
```diff
@@ -35,6 +35,8 @@ use crate::{
     Error, Interface,
 };
 
+/// Interfaces struct
+/// Public in a private module that is never exported
```
joshuachp, line 39:
```suggestion
```

joshuachp (changes requested):

src/store/wrapper.rs
```diff
@@ -45,6 +47,7 @@ where
     }
 }
 
+// TODO could this be a blanket implementation for every S that returns dyn error
```
joshuachp, line 50:
```suggestion
```

The error could be re-worked if we change the Store trait. I don't think an extension trait would be needed though.

src/store/mod.rs
```diff
@@ -49,6 +50,53 @@ pub trait StoreCapabilities {
     fn get_retention(&self) -> Option<&Self::Retention>;
 }
 
+/// Data passed to the store that identifies an interface
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct InterfaceInfo<S = String> {
+    pub(crate) name: S,
+    pub(crate) ownership: Ownership,
+}
```
joshuachp, line 58: Why do we need this to be a String and not an &str? Since it will not be modified.

joshuachp: It will also simplify all the generics in the trait

src/transport/grpc/mod.rs
```diff
@@ -622,13 +632,260 @@ impl<'a> TryFrom<&'a Interfaces> for NodeData {
     }
 }
 
+/// Store implementation designed specifically for the grpc connection
```
joshuachp, line 635: We should separate this into it's own store module

src/transport/grpc/mod.rs
```diff
@@ -622,13 +632,260 @@ impl<'a> TryFrom<&'a Interfaces> for NodeData {
...
+/// Used mainly to request device owned properties to the message hub instead of looking them up in the local storage
+#[derive(Debug, Clone)]
+pub struct GrpcStore<S> {
+    pub(crate) store: S,
+    pub(crate) client: Arc<Mutex<MsgHubClient>>,
```
joshuachp, line 640:
```suggestion
    pub(crate) client: MsgHubClient,
```

joshuachp: I think you can achieve this by modifying the store trait. Since it's Clone and should be accessed by one instance exclusively.

joshuachp (changes requested):

src/store/mod.rs
lucaato: If it was a &str it would clutter some struct it is included in. For example look at `ClientMessage` in `src\connection.rs`. It would be messier to manage. We need it in the Unset variant because then we need to access the store. Are you suggesting that, since unsets should only be performed on device owned interfaces, we just assume it's a device interface ? To avoid having the InterfaceInfo struct inside the ClientMessage struct

src/transport/grpc/mod.rs
lucaato: I'll try this, if it requires too many changes this should be done in another pr.

lucaato: To do this i would need to modify the public interface of the sdk to accept an &mut self or wrap the store in a mutex from the outside, I don't think this should be done in this pr.

src/transport/grpc/store.rs
```diff
@@ -0,0 +1,464 @@
+/*
+ * This file is part of Astarte.
+ *
+ * Copyright 2023 SECO Mind Srl
```
rgallor, line 4: since you are introducing this file, I'd change the year accordingly

src/transport/grpc/mod.rs
joshuachp: You're right, after a more careful consideration I don't think it's worth to change all the client for this change. 

I think we should reconsider this when we want to move the sender to the client.

src/validate.rs
```diff
@@ -196,7 +197,7 @@ impl ValidatedObject {
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub(crate) struct ValidatedUnset {
-    pub(crate) interface: String,
+    pub(crate) interface: InterfaceInfo,
```
joshuachp, line 200:
```suggestion
    pub(crate) interface: String,
```

joshuachp: This is only sent by the device

joshuachp (changes requested):

src/store/mod.rs
lucaato: Removed the generic parameter

src/validate.rs
lucaato: In the end I created a From<ValidatedUnset> for Interface info so that the interface of the store can use it consistently in every method

src/store/mod.rs
```diff
@@ -49,6 +50,43 @@ pub trait StoreCapabilities {
     fn get_retention(&self) -> Option<&Self::Retention>;
 }
 
+/// Data passed to the store that identifies an interface
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct InterfaceInfo<'a> {
+    pub(crate) name: &'a str,
+    pub(crate) ownership: Ownership,
+}
```
joshuachp, line 58: We should add a `PropertyStore` struct to make it simpler to add more arguments to the store functions. It should also implement `Deref` with target `InterfaceInfo` and have public getters for all the fields.

joshuachp (changes requested):

src/store/mod.rs
lucaato: I've implemented the `PropertyInfo` struct, for now I've left out the major version of the interface since in the store it is used in only one method, it can be added in the feature.

joshuachp: I would still add the property info to the struct, otherwise it defeats the purpose of adding arguments without breaking compatibility.

src/store/mod.rs
```diff
@@ -49,6 +51,109 @@ pub trait StoreCapabilities {
     fn get_retention(&self) -> Option<&Self::Retention>;
 }
 
+/// Data passed to the store that identifies a property
+pub struct PropertyInfo<'a> {
+    pub(crate) interface_info: InterfaceInfo<'a>,
+    pub(crate) path: &'a str,
```
joshuachp, line 57:
```suggestion
    interface_info: InterfaceInfo<'a>,
    path: &'a str,
```

src/store/mod.rs
```diff
@@ -49,6 +51,109 @@ pub trait StoreCapabilities {
...
+    fn from(value: &'a ValidatedUnset) -> Self {
+        Self::new_unchecked(value.into(), &value.path)
+    }
+}
+
+/// Data passed to the store that identifies an interface
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct InterfaceInfo<'a> {
+    pub(crate) name: &'a str,
+    pub(crate) ownership: Ownership,
```
joshuachp, line 114:
```suggestion
    name: &'a str,
    ownership: Ownership,
```

joshuachp (changes requested):

src/store/mod.rs
joshuachp: So we are forced to use the public methods in the SDK

src/store/mod.rs
joshuachp: And what do we think of calling them PropPath and PropInterface

rgallor: Yes, `PropertyInterface`  (or `PropInterface`) sounds good to me

joshuachp: For me it's the same 👍🏻

src/transport/grpc/store.rs
```diff
@@ -0,0 +1,444 @@
...
+    S: PropertyStore,
+    S::Err: std::fmt::Display,
+{
+    type Err = GrpcStoreError<S::Err>;
+
+    async fn store_prop(&self, prop: StoredProp<&str, &AstarteType>) -> Result<(), Self::Err> {
+        if prop.ownership.is_device() {
+            // do not store device owned properties
+            return Ok(());
+        }
```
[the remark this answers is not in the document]
lucaato: Until we have a way for the message hub to avoid client overriding each other properties we want to return only the properties of the message hub which are guaranteed to be updated if updated by another client.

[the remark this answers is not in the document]
joshuachp: In the current state, since we unfortunately have no way to prevent the properties to be overwritten I propose the following to prevent them to be lost when disconnected.

- Wen a property is set/unset/deleted we also make the change to the Store
- When we load a property we must ask it to the Message Hub because it can be set by other devices/applications
- When reconnecting we send all the Device properties in the store

This way we do not lose the data that we tried to send while offline. What do you think @harlem88

src/transport/grpc/convert.rs
```diff
@@ -47,7 +49,7 @@ use super::{GrpcError, GrpcPayload};
 
 /// Error returned by the Message Hub types conversions.
 #[non_exhaustive]
-#[derive(thiserror::Error, Debug)]
+#[derive(thiserror::Error, Debug, Clone)]
```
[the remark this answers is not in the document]
lucaato: I removed it

src/transport/grpc/store.rs
```diff
@@ -0,0 +1,444 @@
...
+use super::{
+    convert::{self, MessageHubProtoError},
+    MsgHubClient,
+};
+
+// Store implementation designed specifically for the grpc connection
+/// Used mainly to request device owned properties to the message hub instead of looking them up in the local storage
+#[derive(Debug, Clone)]
+pub struct GrpcStore<S> {
+    pub(crate) store: S,
```
joshuachp, line 49:
```suggestion
```

src/transport/grpc/store.rs
```diff
@@ -0,0 +1,444 @@
...
+    async fn store_prop(&self, prop: StoredProp<&str, &AstarteType>) -> Result<(), Self::Err> {
+        if prop.ownership.is_device() {
+            // do not store device owned properties
+            return Ok(());
+        }
+
+        self.store
+            .store_prop(prop)
+            .await
+            .map_err(GrpcStoreError::from_err)
```
joshuachp, line 141:
```suggestion
    Ok(())
```

joshuachp: The property should be stored on the message hub

src/transport/grpc/store.rs
```diff
@@ -0,0 +1,444 @@
...
+        property: &PropertyMapping<'_>,
+        interface_major: i32,
+    ) -> Result<Option<AstarteType>, Self::Err> {
+        if property.ownership().is_server() {
+            return self
+                .store
+                .load_prop(property, interface_major)
+                .await
+                .map_err(GrpcStoreError::from_err);
+        }
```
joshuachp, line 155:
```suggestion
```
Same as above, as discussed we should ask the Message Hub

src/transport/grpc/convert.rs
```diff
@@ -60,6 +62,72 @@ pub enum MessageHubProtoError {
...
+                    value,
+                    interface_major: prop_data.version_major,
+                    ownership: prop_data.ownership().into(),
+                };
+
+                Ok(res)
+            })
+        })
+        .try_collect()
+}
```
joshuachp, line 112: Do we want to error on unset in the store, or should we filter them?

joshuachp (changes requested):

src/validate.rs
```diff
@@ -219,6 +227,13 @@ impl ValidatedUnset {
     }
 }
 
+impl<'a> From<&'a ValidatedUnset> for PropertyInterface<'a> {
+    /// an unset is allowed to be sent only by the device [`ValidatedUnset::validate`]
+    fn from(value: &'a ValidatedUnset) -> Self {
+        Self::new(&value.interface, Ownership::Device)
+    }
+}
```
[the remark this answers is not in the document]
lucaato: I wanted to keep it near the validate method to show clearly that the constructor only allows device properties

src/transport/grpc/convert.rs
lucaato: This was something we discussed and we decided that the message hub should not return unset values if I'm not mistaken. This is probably something that from the server side handles too, @rgallor do you remember something similar ?

src/transport/grpc/store.rs
```diff
@@ -0,0 +1,353 @@
...
+    Status(tonic::Status),
+    /// Error while converting a proto received value to an internal type
+    #[error("Error while converting a proto received value to an internal type: {0}")]
+    Conversion(MessageHubProtoError),
+}
+
+impl GrpcStoreError {
+    fn from_status(status: tonic::Status) -> Self {
+        Self::Status(status)
+    }
```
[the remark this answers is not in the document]
lucaato: Used a from impl

src/transport/grpc/convert.rs
lucaato: makes sense

joshuachp (approved):

src/builder.rs
```diff
@@ -311,96 +272,151 @@ where
...
 }
 
-impl<S, C> Debug for DeviceBuilder<S, C>
+impl<S, C> DeviceBuilder<S, C>
 where
-    C: Connection,
+    S: PropertyStore,
+    C: ConnectionConfig<S>,
+    crate::Error: From<C::Err>,
+    C: RetentionOverrideConfig,
```
lucaato, line 328: Added this trait bound and trait to allow the grpc connection to disable the volatile retention by setting the capacity to `0`.

src/builder.rs
```diff
@@ -311,96 +272,151 @@ where
...
+        }
+    }
+}
+
+// NOTE: Currently volatile retention is not supported for the Grpc connection.
+// Currently messages won't be retained or stored (see [`crate::transport::store::GrpcStore`])
+impl<S> DeviceBuilder<S, crate::transport::mqtt::MqttConfig> {
+    /// Sets the number of elements with retention volatile that will be kept in memory if
+    /// disconnected.
+    pub fn volatile_retention(mut self, items: usize) -> Self {
```
lucaato, line 316: I moved this method here so that the user can call it only after setting the connection config to `MqttConfig`. She won't be able to call this method when the connection config is `GrpcConfig` as the config would be overridden anyway.

src/retention/memory.rs
```diff
@@ -92,7 +92,8 @@ struct VolatileStore {
 impl VolatileStore {
     fn new() -> Self {
         Self {
-            store: VecDeque::new(),
+            // we use 1 as default to allow the store of elements
+            store: VecDeque::with_capacity(1),
```
joshuachp, line 96: I don't think this is necessary or too good for the store, if we create a new `VolatileStore` without capacity we don't want to allocate nothing and then set the appropriate capacity

joshuachp (changes requested):

src/retention/memory.rs
joshuachp: Furthermore, having just one element is not that useful for the store, but it still needs to be allocated.

rgallor (changes requested):

src/retention/memory.rs
lucaato: Ok

src/builder.rs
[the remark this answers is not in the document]
rgallor: make sense!

src/retention/memory.rs
```diff
@@ -551,4 +558,26 @@ mod tests {
...
+            version_major: 0,
+            reliability: Reliability::Unique,
+            retention: Retention::Volatile { expiry: None },
+            data: AstarteType::Integer(42),
+            timestamp: None,
+        };
+
+        store.push(ctx.next(), info.clone());
+        store.push(ctx.next(), info.clone());
+        store.push(ctx.next(), info.clone());
```
rgallor, line 578: are all of them necessary for the test?

src/store/sqlite/mod.rs
```diff
@@ -524,12 +537,18 @@ impl PropertyStore for SqliteStore {
...
+        self.with_reader(|reader| reader.interface_props(interface.name.as_ref()))
     }
 
-    async fn delete_interface(&self, interface: &str) -> Result<(), Self::Err> {
-        self.writer.lock().await.delete_interface_props(interface)?;
+    async fn delete_interface(&self, interface: &PropertyInterface<'_>) -> Result<(), Self::Err> {
+        self.writer
+            .lock()
+            .await
+            .delete_interface_props(interface.name.as_ref())?;
```
rgallor, line 551: you could directly use `interface.name` or `interface.name()` rather than calling `as_ref()`

src/store/sqlite/mod.rs
```diff
@@ -524,12 +537,18 @@ impl PropertyStore for SqliteStore {
         self.with_reader(|reader| reader.props_with_ownership(Ownership::Server))
     }
 
-    async fn interface_props(&self, interface: &str) -> Result<Vec<StoredProp>, Self::Err> {
-        self.with_reader(|reader| reader.interface_props(interface))
+    async fn interface_props(
+        &self,
+        interface: &PropertyInterface<'_>,
+    ) -> Result<Vec<StoredProp>, Self::Err> {
+        self.with_reader(|reader| reader.interface_props(interface.name.as_ref()))
```
rgallor, line 544: same here

rgallor, line 544: same here

src/store/sqlite/mod.rs
```diff
@@ -494,14 +501,20 @@ impl PropertyStore for SqliteStore {
...
         Ok(())
     }
 
-    async fn delete_prop(&self, interface: &str, path: &str) -> Result<(), Self::Err> {
-        self.writer.lock().await.delete_prop(interface, path)?;
+    async fn delete_prop(&self, property: &PropertyMapping<'_>) -> Result<(), Self::Err> {
+        self.writer
+            .lock()
+            .await
+            .delete_prop(property.name.as_ref(), property.path)?;
```
rgallor, line 517: also here

src/store/sqlite/mod.rs
```diff
@@ -494,14 +501,20 @@ impl PropertyStore for SqliteStore {
         }
     }
 
-    async fn unset_prop(&self, interface: &str, path: &str) -> Result<(), Self::Err> {
-        self.writer.lock().await.unset_prop(interface, path)?;
+    async fn unset_prop(&self, property: &PropertyMapping<'_>) -> Result<(), Self::Err> {
+        self.writer
+            .lock()
+            .await
+            .unset_prop(property.name.as_ref(), property.path)?;
```
rgallor, line 508: also here

src/store/sqlite/mod.rs
```diff
@@ -466,24 +468,29 @@ impl PropertyStore for SqliteStore {
...
     async fn load_prop(
         &self,
-        interface: &str,
-        path: &str,
+        property: &PropertyMapping<'_>,
         interface_major: i32,
     ) -> Result<Option<AstarteType>, Self::Err> {
-        let opt_record = self.with_reader(|reader| reader.load_prop(interface, path))?;
+        let opt_record =
+            self.with_reader(|reader| reader.load_prop(property.name.as_ref(), property.path))?;
```
rgallor, line 475: also here

rgallor (changes requested):

src/retention/memory.rs
```diff
@@ -108,6 +109,11 @@ impl VolatileStore {
         T: TryInto<ItemValue, Error = VolatileItemError>,
     {
         if self.is_full() {
+            if self.store.capacity() == 0 {
+                // we shouldn't store anything
+                return;
+            }
```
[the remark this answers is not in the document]
lucaato: It was to avoid having multiple if branching the main path... I don't think it changes much either way but if it is not full we already know we can push an element so I wouldn't move it.

src/retention/memory.rs
lucaato: It is just to show that no amounts of pushing will have effect but it's the same

src/store/mod.rs
```diff
@@ -49,6 +51,109 @@ pub trait StoreCapabilities {
...
+            path,
+        }
+    }
+
+    pub(crate) fn new_unchecked(interface_info: PropertyInterface<'a>, path: &'a str) -> Self {
+        Self {
+            interface_info,
+            path,
+        }
+    }
```
rgallor, line 73: I would use a single method like the following:


```suggestion
    pub(crate) fn new_unchecked(interface_info: impl Into<PropertyInterface<'a>>, path: &'a str) -> Self {
        Self {
            interface_info: interface_info.into(),
            path,
        }
    }
```

lucaato: No, I'd prefer to keep them separated since `new_unchecked` could be considered an `unsafe` method because you have to make sure that the interface you are passing is actually a property interface but we have no way of knowing. At the same time `with_property_path` is checked by our other methods to receive an input property interface.
This is only used inside the crate so it's not a big deal, but it would be better to keep them separated IMO. If this doesn't make sense to most of us I can remove it.

src/transport/grpc/store.rs
```diff
@@ -0,0 +1,345 @@
...
+impl StoreCapabilities for GrpcStore {
+    type Retention = Missing;
+
+    fn get_retention(&self) -> Option<&Self::Retention> {
+        None
+    }
+}
+
+/// We implement the PropertyStore to override the behavior when retrieving or storing
+/// device owned properies. We only want to store and load from the store server owned properties.
```
harlem88, line 106: Nit: Change to `properties`
```suggestion
/// device owned properties. We only want to store and load from the store server owned properties.
```

harlem88 (changes requested):
LGTM

harlem88 (approved):

PR #403, opened 2025-01-09
chore(deps): update rumqttc dependency

joshuachp:
Fixes an upstream bug in the MQTT unsubscribe.

### DIFF
--- a/src/transport/mqtt/config/tls.rs
+++ b/src/transport/mqtt/config/tls.rs
@@ -29,7 +29,7 @@ use rustls::{
     pki_types::{CertificateDer, PrivatePkcs8KeyDer},
     RootCertStore,
 };
-use tracing::{debug, error, warn};
+use tracing::{debug, error, trace, warn};
 
 use crate::transport::mqtt::PairingError;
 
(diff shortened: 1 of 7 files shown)


rgallor (approved):

src/transport/mqtt/config/tls.rs
```diff
@@ -131,13 +131,16 @@ async fn read_root_cert_store() -> Result<RootCertStore, PairingError> {
...
 
-        for cert in native_certs {
-            root_cert_store.add(cert).map_err(PairingError::Tls)?;
+        for err in res.errors {
+            error!(error = %crate::error::Report::new(err), "couldn't load root certificate");
         }
 
+        let (added, ignored) = root_cert_store.add_parsable_certificates(res.certs);
+
+        debug!("loaded {added} certs and {ignored} ignored");
```
harlem88, line 142: Do we want to add these logs?
```suggestion
```

joshuachp: We could make it a trace, but it should only be printed once at startup

lucaato (approved):

harlem88 (approved):

PR #409, opened 2025-01-24
ci(docs): include documentation in for publish

joshuachp:


harlem88 (approved):

scripts/check-for-publish.sh
```diff
@@ -0,0 +1,46 @@
...
+# List files in a package
+listPackage() {
+    cargo package -l -p "$1" | xargs -I '{}' echo "$1/{}"
+}
+
+pkgsFiles=$(
+    cat <(cargo package -l -p "astarte-device-sdk") \
+        <(listPackage "astarte-device-sdk-derive") \
+        <(listPackage "astarte-device-sdk-mock") \
+        <(listPackage "e2e-test") |
```
harlem88, line 34: Are we need to test also `e2e-test` for publishing?
```suggestion
```

harlem88 (changes requested):

scripts/check-for-publish.sh
joshuachp: Yes, I tried without, but since it's a member of the workspace we need to check its files too.

joshuachp:
I fixed an error in the comment and re-pushed

harlem88 (approved):

PR #413, opened 2025-02-13
fix(props): filter properties that are actually in the introspection 

joshuachp:
Retain only the properties that are currently in the device
introspection, this fixes an issue when during an update an interface is
removed. This should be further improved by keeping track of the
introspection of the device.


harlem88 (commented):
Add this bugfix to the changelog

harlem88 (approved):

PR #423, opened 2025-03-03
fix(rumqttc): use the new version of rumqttc

joshuachp:
Fix a concurrency bug in the task that waits for the Acks from astarte and the one that handles the event loop.

By updating the rumqttc crate other dependencies have been bumped, mainly rustls. It introduced an error in some tests that generated the CSR for the Astarte Certificates. Now a self signed CSR/Certificate is required in the tests (to pass to rustls). This has made it hard to maintain openssl, since it generates a non commonly used certificate. I decided to remove it as a dependency, since we are fully using rustls anyway and it was only for backwards compatibility.

See comment https://github.com/astarte-platform/astarte-device-sdk-rust/pull/423#discussion_r1979652570 for a better explenation


src/transport/mqtt/connection.rs
```diff
@@ -661,25 +661,30 @@ impl WaitAcks {
...
+                let next = match res {
+                    Ok(event) => Next::handle_event(event),
+                    Err(err) => Next::handle_error(err),
+                };
 
-        let next = match conn.eventloop_mut().poll().await {
-            Ok(event) => Next::handle_event(event),
-            Err(err) => Next::handle_error(err),
+                return Ok(next)
+            }
```
joshuachp, line 681: This is where the bug was, now we use the select to ovviate the problem

joshuachp:
I notice that we don't pass the `ignore_ssl` to the pairing client. For now, I added the `reqwest` feature to load native certs to pass the CI.

src/transport/mqtt/crypto.rs
```diff
@@ -78,7 +71,7 @@ impl Bundle {
         dn.push(DnType::CommonName, format!("{}/{}", realm, device_id));
 
         // Generate a random private key
-        let key_pair = KeyPair::generate_for(&PKCS_ECDSA_P384_SHA384)?;
+        let key_pair = KeyPair::generate_for(&PKCS_ECDSA_P256_SHA256)?;
```
harlem88, line 74: it is not a problem, but why do you decrement curve and SHA value to `256`? Did you see a performance problem?

harlem88 (changes requested):

src/transport/mqtt/crypto.rs
joshuachp: I changed it while tasting the self signing of the CSR

https://github.com/astarte-platform/astarte-device-sdk-rust/blob/0818d353bd45a349c0fbe7bb28bb50bc346fb817/src/transport/mqtt/pairing.rs#L292-L300

The validation of the CSR was failing, so I wanted a more common certificate. In the end it was the OpenSSL one that was really particular and not supported by rcgen/rustls (for the self signing and validation of the CSR). So I end up removing OpenSSL, since it wasn't actually used in the final binaries.

I think we can use the PKCS_ECDSA_P384_SHA384 private key type without problems. Still  we can consider moving to the PKCS_ECDSA_P256_SHA256, since it's the one suggested by `rcgen/rustls` and we don't have a reason to use the PKCS_ECDSA_P384_SHA384.

joshuachp: The `self_sign_csr_to_pem` uses the key generated by `astarte-cfssl-config`, and it's the PKCS_ECDSA_P256_SHA256 one.

joshuachp: The old certificate would still work, and I tested:

- running the e2e-tests on release-0.9, which generates the old cert and private key in the store directory
- Move to the PR branch and re-launch the e2e-tests using the same certificate and key without a problem

The new one will be re-generate when the old one expires or the store-directory is cleaned, or the device is factory reset

rgallor (approved):

src/transport/mqtt/connection.rs
```diff
@@ -655,31 +655,30 @@ impl WaitAcks {
...
+            res = &mut self.handle => {
+                debug!("task joined");
 
-            return self.handle_join().await;
-        }
-
-        trace!("polling next event");
+                Self::handle_join(res)
+            }
+            // I hope this is cancel safe
```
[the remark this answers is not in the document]
joshuachp: Pray the 🦀

harlem88 (approved):
LGTM. Add a line to the changelog, about these changes

src/transport/mqtt/connection.rs
```diff
@@ -655,31 +655,30 @@ impl WaitAcks {
...
     async fn wait_connection(&mut self, conn: &mut Connection) -> Result<Next, StoreError> {
-        // HACK: This is to make the function pass with mockall, since the poll function would never
-        //       yield while using the mocks. With the real implementation the `Eventloop::poll`
-        //       would have to wait for the connection and the sent packets.
-        #[cfg(test)]
-        tokio::task::yield_now().await;
-
-        if self.handle.is_finished() {
-            debug!("init task finished");
+        tokio::select! {
```
lucaato, line 658: Why a simple await on the handle join would not suffice anymore?
Why did you move to `tokio::select`?
Did you consider using something like [`futures::future::select`](https://docs.rs/futures/latest/futures/future/fn.select.html) which lets you handle cancellation explicitly?

src/builder.rs
```diff
@@ -615,12 +619,25 @@ mod test {
...
+                        tokio::task::yield_now().await;
+
+                        Ok(Event::Incoming(Packet::ConnAck(ConnAck {
+                            session_present: false,
+                            code: ConnectReturnCode::Success,
+                        })))
+                    })
                 });
 
+            // catch other calls to pool
```
lucaato, line 632: I see this catchall expect got added in other tests too, is it because of the select in the `WaitAcks` which makes it more unpredictable ?

src/transport/mqtt/config/transport.rs
```diff
@@ -288,6 +303,11 @@ mod tests {
 
     #[tokio::test]
     async fn should_recreate_transport() {
+        tracing_subscriber::fmt()
+            .with_test_writer()
+            .with_max_level(tracing::Level::TRACE)
+            .init();
```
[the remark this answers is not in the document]
joshuachp: Used during tests, removed thanks

src/builder.rs
joshuachp: Before we used a HACK in the wait_connection method, the gist of it is that we are not guarantied which future is polled by the Tokio runtime (it's random by design). To solve this, we return a new future that will never complete.

src/transport/mqtt/connection.rs
lucaato: Thank you for the thorough explanation :heart_hands:

lucaato (approved):

harlem88 (approved):

PR #425, opened 2025-03-05
feat(crypto): use default crypto provider

rgallor:
Set default `CryptoProvider` to avoid application errors


src/transport/mqtt/config/tls.rs
```diff
@@ -107,10 +108,17 @@ impl ClientAuth {
         Ok(Some(ClientAuth { private_key, certs }))
     }
 
-    pub(crate) async fn tls_config(self) -> Result<rustls::ClientConfig, PairingError> {
-        let roots = read_root_cert_store().await?;
-
-        rustls::ClientConfig::builder()
+    pub(crate) async fn tls_config(
```
joshuachp, line 111:
```suggestion
    pub(crate) fn tls_config(
```

src/transport/mqtt/config/tls.rs
```diff
@@ -107,10 +108,17 @@ impl ClientAuth {
...
+        self,
+        roots: Arc<RootCertStore>,
+    ) -> Result<rustls::ClientConfig, PairingError> {
+        let crypto_provider = CryptoProvider::get_default()
+            .cloned()
+            .unwrap_or_else(|| Arc::new(rustls::crypto::aws_lc_rs::default_provider()));
+
+        rustls::ClientConfig::builder_with_provider(crypto_provider)
+            .with_safe_default_protocol_versions()
```
joshuachp, line 120: Separate this into a function (not a method of client auth) to share with the TransportProvider:

```rust
pub(crate) fn tls_client_config(roots: Arc<RootCertStore>) -> Result<rustls::ConfigBuilder<ClientConfig, WantsCerts>, PairingError> {
     let crypto_provider = CryptoProvider::get_default()
            .cloned()
            .unwrap_or_else(|| Arc::new(rustls::crypto::aws_lc_rs::default_provider()));
    
     
        rustls::ClientConfig::builder_with_provider(crypto_provider)
            .with_safe_default_protocol_versions()
            .map_err(PairingError::Tls)?
            .with_root_certificates(roots)
}
```

joshuachp: Also  the insecure one need to be added the same way. Additionally passing the CrytpoProvider to that too

joshuachp (changes requested):

src/transport/mqtt/config/tls.rs
```diff
@@ -127,27 +135,30 @@ impl ClientAuth {
...
+        root_cert_store
     })
-    .await?
+    .await
+    .map_err(PairingError::ReadNativeCerts)
 }
 
 #[derive(Debug)]
-struct NoVerifier;
+pub(crate) struct NoVerifier;
```
joshuachp, line 161:
```suggestion
struct NoVerifier;
```

src/transport/mqtt/config/transport.rs
```diff
@@ -38,21 +41,48 @@ pub(crate) struct TransportProvider {
...
+        } else {
+            let crypto_provider = CryptoProvider::get_default()
+                .cloned()
+                .unwrap_or_else(|| Arc::new(rustls::crypto::aws_lc_rs::default_provider()));
+
+            rustls::ClientConfig::builder_with_provider(crypto_provider)
+                .with_safe_default_protocol_versions()
+                .map_err(PairingError::Tls)?
+                .with_root_certificates(Arc::clone(&self.root_cert_store))
+                .with_no_client_auth()
```
joshuachp, line 80: Here this will go into the tls module with the new functions

joshuachp (changes requested):

src/transport/mqtt/pairing.rs
```diff
@@ -169,7 +147,10 @@ impl<'a> ApiClient<'a> {
 
         let payload = ApiData::new(MqttV1Csr { csr });
 
-        let client = reqwest::Client::new();
+        let client = reqwest::Client::builder()
+            .use_preconfigured_tls(self.tls_config.clone())
+            .build()?;
```
joshuachp, line 152: We can configure the client once and put it in the Api struct

joshuachp (approved):

Issue #412, opened 2025-02-07
[API]: Unify Send API


harlem88:
Currently each Astarte Device SDKs has a different name for the function that allows publishing data to an Astarte instance. 

We want to unify the Send API across all SDKs to avoid confusion for the end user, who may use different SDKs for various use cases and could become disoriented.

All SDKs will implement the unified API for publishing data to an Astarte instance.

API for publishing a property value :

``` rust
fn set_property(...);

fn unset_property(...);
```

API for publishing a individual value :

``` rust
fn send_individual(...);
```

API for publishing a object value :

``` rust
fn send_object(...);
```

PR #430, opened 2025-03-19
refactor(object)!: remove the trait AstarteAgregate for the struct AstarteObject and macro IntoAstarteObject

joshuachp:
This will conform to the changes on the API side.

This is in preparation of https://github.com/astarte-platform/astarte-device-sdk-rust/issues/412


joshuachp:
I've added a new struct `AstarteObject` to simplify the public API for the send methods. The trait bounds are no longer required, while the macro is still present.

rgallor (approved):

src/aggregate.rs
```diff
@@ -18,53 +18,123 @@
...
+    /// Returns a reference to the value with the given name, if present.
+    pub fn get(&self, name: &str) -> Option<&AstarteType> {
+        self.inner
+            .iter()
+            .find_map(|(item, value)| (*item == name).then_some(value))
+    }
+
+    /// Remove the value with the given name, if present.
+    pub fn remove(&mut self, name: &str) -> Option<AstarteType> {
+        let (position, _) = self.inner.iter().find_position(|(item, _)| *item == name)?;
```
lucaato, line 91: Could you use the std `position` method and achieve the same result ?

src/client.rs
```diff
@@ -410,9 +405,7 @@ impl<S> DeviceClient<S> {
 
         debug!("sending {} {}", interface_name, path);
 
-        let aggregate = data.astarte_aggregate()?;
-
-        let validated = ValidatedObject::validate(object, path, aggregate, timestamp)?;
+        let validated = ValidatedObject::validate(object, path, data, timestamp)?;
```
lucaato, line 408: nit: now that the struct is called `AstarteObject` i would rename the `object` variable to something like `interface`

src/aggregate.rs
```diff
@@ -18,53 +18,123 @@
...
-    ///         let mut r = HashMap::new();
-    ///         r.insert("name".to_string(), self.name.try_into()?);
-    ///         r.insert("age".to_string(), self.age.try_into()?);
-    ///         r.insert("phones".to_string(), self.phones.try_into()?);
-    ///         Ok(r)
-    ///     }
-    /// }
-    /// ```
-    fn astarte_aggregate(self) -> Result<HashMap<String, AstarteType>, Error>;
+    /// The value are unique and ordered by insertion. It overrides and returns the previous value
```
lucaato, line 69:
```suggestion
    /// The values are unique and ordered by insertion. It overrides and returns the previous value
```
nit

src/aggregate.rs
```diff
@@ -1,12 +1,12 @@
 // This file is part of Astarte.
```
lucaato, line 1: Should we rename the module too ?

lucaato (commented):
Mostly nitpicking gj

lucaato (approved):

src/aggregate.rs
```diff
@@ -18,53 +18,133 @@
...
+/// let mut object = AstarteObject::new();
+/// object.insert("name".to_string(), sensor.clone());
+/// object.insert("id".to_string(),  id.clone());
+///
+/// assert_eq!(object.get("name"), Some(&sensor));
+/// assert_eq!(object.get("id"), Some(&id));
+/// ```
+#[derive(Debug, Clone, Default)]
+pub struct AstarteObject {
+    pub(crate) inner: Vec<(String, AstarteType)>,
```
harlem88, line 50: Why we changed the Map in favor of Vec to manage the object fields? I noted also the introduction of some methods to find/put/get/remove that are already defined in a map.

joshuachp: The motivation are the following:

- The std hash map in rust is pretty costly since it's a general purpose once and by default calls get random and uses a cryptographically secure hash function to maximize DDoS protection. Since the user has control of the inputs, those guaranties only hinder the program[^1]. 
- Also, a hash map for a low number of keys, which is the usual case for the Astarte object, would require to allocate a larger capacity to have less collision or use linear probing. When inserting, all the keys need to be also rehashed when resizing the hash map. So the cost of the hash map is not amortized.
- Using a better hashing function would require adding a new dependency like [rust-c-hash](https://crates.io/crates/rustc-hash) which I have considered instead of using a vector.
- I did not benchmark the change, but doing a linear searching for a key in the with a low number of elements (like in the case of the Astarte Object) is __usually__ comparable to accessing an HashMap.
- We also don't do random access of the AstarteObject in the SDK, we only use it as a DTO to serialize it and send it to Astarte. We iterate over its key values for validation[^2], so a Vec is more suited for this behavior. The user would only insert the key-values in the AstarteObject or use the derive macro to convert into it.

The focus of the change was more to mask the underling data structure of the `AstarteObject` so we can still change it in the future without making a breaking change.

IMO, a Vec is better suited for our use case

[^1]: https://nnethercote.github.io/perf-book/hashing.html
[^2]: https://github.com/joshuachp/astarte-device-sdk-rust/blob/1d603512a526dce251800888acd6c8255418326a/src/validate.rs#L163

harlem88: Thanks! All valid reasons. I agree with you, and I resolve the comment. Great Job!

harlem88 (approved):

PR #431, opened 2025-03-20
refactor!(validation): make it stricter to reflect the new API

joshuachp:
Check the interface type, ownership and aggregation in the user
validation. Return an hard error if one is incorrect.

### DIFF
--- a/e2e-test/src/main.rs
+++ b/e2e-test/src/main.rs
@@ -35,6 +35,7 @@ use std::time::Duration;
 
 use astarte_device_sdk::transport::mqtt::Credential;
 use astarte_device_sdk::{Interface, Value};
+use chrono::Utc;
 use eyre::{bail, eyre, OptionExt, WrapErr};
 use itertools::Itertools;
 use reqwest::StatusCode;
@@ -140,8 +141,8 @@ async fn main() -> eyre::Result<()> {
         .with(EnvFilter::from_default_env())
         .try_init()?;
 
-    let default_provider = rustls::crypto::aws_lc_rs::default_provider();
-    rustls::crypto::CryptoProvider::install_default(default_provider)
+    rustls::crypto::aws_lc_rs::default_provider()
+        .install_default()
         .map_err(|_| eyre!("couldn't install default crypto provider"))?;
 
     let test_cfg = TestCfg::init().wrap_err("Failed configuration initialization")?;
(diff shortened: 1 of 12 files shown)


joshuachp:
This PR doesn't change the logic of the explicit timestamp. I wonder if we should do the following:

- If explicit_timestamp is set, require a timestamp to be send
  - Currently this is not an error, and astarte accepts it
- If explisic_timestamp is not set, and a timestamp is sent, return an error
  - This is currently enforced

IMO, this shouldn't be a server error, but should be enforced in the client.

joshuachp:
After a talk with the cloud backend team, we agreed that it's beneficial to add the check to the client SDK. Since this is a major version bump, I've also added this to the breaking changes for review.

harlem88 (approved):

rgallor (approved):

src/lib.rs
```diff
@@ -119,6 +119,19 @@ mod test {
...
+    "version_major": 0,
+    "version_minor": 1,
+    "type": "properties",
+    "ownership": "device",
+    "mappings": [{
+        "endpoint": "/%{sensor_id}/enable",
+        "type": "boolean",
+        "allow_unset": false
+    }]
+}"#;
```
lucaato, line 133: Since other interfaces are included as files would it make sense to include this one too as a file ?

e2e-test/src/main.rs
```diff
@@ -409,10 +415,11 @@ async fn test_aggregate_device_to_server(
     // Send all the mock test data
     debug!("Sending device owned aggregate from device to server");
     device
-        .send_object(
+        .send_object_with_timestamp(
             &test_cfg.interface_aggregate_do,
             &format!("/{sensor_number}"),
             tx_data.clone().try_into().unwrap(),
+            Utc::now(),
```
lucaato, line 422: In the next pr about the e2e refactor it would be cool to also check the timestamp as received by astarte if possibile.

lucaato (commented):
Just some nits and questions

e2e-test/src/main.rs
joshuachp: We cannot, Astarte will never send a timestamp to the device for now.

joshuachp: But I have added more logic for the received timestamp in the refactor of the API

src/lib.rs
joshuachp: Where do you think I should move it to? I think it's easier to add test with inline interfaces (with no better alternative). But if you want to propose a better place, I could move it there.

e2e-test/src/main.rs
lucaato: > We cannot, Astarte will never send a timestamp to the device for now.

No but we send the timestamp to the server, and we can check that the http api (or websocket) return that same timestamp or am i mistaken ?

src/lib.rs
lucaato: If we had a property interface in the e2e with `allow_unset: false` then we could use that one. It's something that could make sense having in the e2e. If it makes sense it could also be done in the next pr.

lucaato (approved):

e2e-test/src/main.rs
joshuachp: I misunderstood, I will address it in the next PR

PR #442, opened 2025-04-10
feat(new-api): refactor the client and connection to follow the new SDK public API

joshuachp:
Closes https://github.com/astarte-platform/astarte-device-sdk-rust/issues/412


lucaato (changes requested):

src/transport/mqtt/config/mod.rs
```diff
@@ -373,22 +372,16 @@ impl MqttConfig {
...
-    ) -> Result<DeviceTransport<Self::Store, Self::Conn>, Self::Err> {
-        let ConnectionBuildConfig {
-            store,
-            interfaces,
-            config,
-        } = config;
-
+        config: BuildConfig<S>,
+    ) -> Result<DeviceTransport<Self::Conn>, Self::Err> {
         let secret = self.credentials(&config).await?;
```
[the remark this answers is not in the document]
joshuachp: I think this is an interesting, and we could secure the secret when in memory. We should open an issue for this, maybe when integrating with the TPM.

src/transport/mqtt/connection.rs
```diff
@@ -199,54 +199,70 @@ impl MqttConnection {
...
             let opt_publish = self
                 .state
                 .poll(&mut self.connection, client_id, interfaces, store)
                 .await?;
 
             if let Some(publish) = opt_publish {
-                debug!("received a publish");
+                debug!("publish received");
 
                 disable_clean_session(self.connection.eventloop_mut());
```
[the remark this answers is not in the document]
joshuachp: I moved, it's now called at least once when we received a connack

examples/individual_properties/main.rs
```diff
@@ -118,10 +118,10 @@ async fn main() -> eyre::Result<()> {
         // Send in a loop the change of the property "name" of sensor 1
         loop {
             client_cl
-                .send(
+                .send_individual(
```
harlem88, line 121: Why don't we use the new property api?
```suggestion
                .set_property(
```

harlem88 (changes requested):

rgallor (changes requested):

rgallor:
LGTM, nice changes 💪🏻

harlem88 (approved):

astarte-device-sdk-derive/src/event.rs
```diff
@@ -227,50 +274,149 @@ impl FromEventDerive {
...
+
+            quote! {
+                Endpoint::<&str>::try_from(#endpoint)?
+            }
+        });
+
         let variants = variants.iter().enumerate().map(|(i, v)| {
             let variant = &v.name;
 
             if v.attrs.allow_unset {
```
lucaato, line 373: Instead of repeating the whole chunk of code you could do this:
```rust
let unset_cond = if v.attrs.allow_unset {
                quote! {
                    Ok(#name::#variant(None))
                }
            } else {
                quote! {
                    return Err(FromEventError::Unset {
                        interface: INTERFACE,
                        endpoint: event.path,
                    });
                }
            };
```
And then use the variable in the other quote block

examples/individual_datastream/main.rs
```diff
@@ -83,22 +84,22 @@ async fn main() -> eyre::Result<()> {
...
                 .await?;
             println!("Data sent on endpoint 1, content: {elapsed}");
             // Sleep 1 sec
             tokio::time::sleep(Duration::from_secs(1)).await;
             // Send endpoint 2
-            let elapsed: f64 = now.elapsed()?.as_secs_f64();
+            let elapsed = now.elapsed()?.as_secs_f64();
             client_cl
-                .send(
+                .send_individual(
```
lucaato, line 99: This is still missing the timestamp, isn't it required in the interface ?

lucaato (changes requested):

examples/individual_datastream/main.rs
lucaato: Maybe in the future we could add a step to the e2e tests that configures and runs the examples to verify they are still working. It could be useful since they are probably the first thing a user tries out.

astarte-device-sdk-derive/src/event.rs
joshuachp: You are missing the `Value::Property(None)` case here

joshuachp: I reduced the code duplication

examples/individual_datastream/main.rs
joshuachp: Fixed and opened https://github.com/astarte-platform/astarte-device-sdk-rust/issues/451

lucaato (approved):

PR #446, opened 2025-05-06
refactor(retry): improve the retry logic 

joshuachp:
Keep track of the instant the last value was generated in the
ExponentialIterator and a max reset duration. This will prevent infinite
loops on immediate reconnection.

### DIFF
--- a/src/connection/resend.rs
+++ b/src/connection/resend.rs
@@ -25,7 +25,6 @@ use crate::builder::DEFAULT_CHANNEL_SIZE;
 use crate::error::Report;
 use crate::retention::memory::ItemValue;
 use crate::retention::{RetentionId, StoredRetention, StoredRetentionExt};
-use crate::retry::ExponentialIter;
 use crate::state::SharedState;
 use crate::store::wrapper::StoreWrapper;
 use crate::store::StoreCapabilities;
@@ -115,7 +114,7 @@ where
 
             match resend.await {
                 Ok(()) => {
-                    trace!("resend task joined")
+                    trace!("task already exited")
                 }
                 Err(err) if err.is_cancelled() => {
                     debug!("resend task was cancelled");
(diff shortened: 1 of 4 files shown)


rgallor (approved):

src/connection/resend.rs
```diff
@@ -191,10 +190,18 @@ where
     {
         let interfaces = self.state.interfaces.read().await;
         debug!("reconnecting");
-        let mut exp = ExponentialIter::default();
+
+        // Wait for the first reconnection to prevent an immediate retry when the reconnection is
+        // actually successful immediately (not entering the while loop). An example is not
+        // installing the interfaces on Astarte.
+        let timeout = self.backoff.next();
```
lucaato, line 197: Instead of storing the backoff struct, wouldn't it be easier just to have the first backoff be `1`? This also would simplify the struct since we would not need a new reset timeout. As per the reset timeout, is it supposed to help when more disconnections happen in a narrow time frame? Honestly, I think it would work as it was handled before too: every time we connect, the exponential retry timer gets reset and starts again from a low timeout.

joshuachp: An initial back off of 1 doesn't help in the case explained in the comment, the reset timeout does.

Think about the case when the interfaces are not installed. We immediately reconnect to Astarte and get booted out when trying to publish the introspection.

lucaato: Maybe the comment should be expanded upon, I don't think that astarte kicks the client out when publishing the introspeciton, that happens when something gets published on an interface that astarte does not have registered, otherwise the reconnection loop would not exit, so yeah it makes sense that we store the exponential retry to at least spam less error messages

joshuachp: Sorry, we get disconnected after re-sending the device properties or publishing new messages immediately after the reconnect.

joshuachp: Tell me if you think the comment can be expanded to be made clearer please

lucaato (approved):

PR #447, opened 2025-05-06
fix(mqtt): use a manual future to poll the retention tokens

joshuachp:
This will make the select more responsive when a future is ready, before
it would take one more call to the poll function.

### DIFF
--- a/src/transport/mqtt/retention.rs
+++ b/src/transport/mqtt/retention.rs
@@ -25,10 +25,12 @@
 //! When an interface major version is updated the retention cache must be invalidated. Since the
 //! payload will be publish on the new introspection.
 
+use std::future::Future;
+use std::pin::Pin;
 use std::{collections::HashMap, future::IntoFuture, task::Poll};
 
 use rumqttc::{AckOfPub, Token, TokenError};
-use tracing::trace;
+use tracing::{trace, warn};
 
 use crate::retention::RetentionId;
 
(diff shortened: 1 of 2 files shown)


rgallor (approved):

src/transport/mqtt/retention.rs
```diff
@@ -104,29 +89,34 @@ impl<'a> IntoFuture for &'a mut MqttRetention {
...
 
-        this.0.queue();
+        let first = this.packets.iter_mut().find_map(|(id, token)| {
+            let poll = <Token<AckOfPub> as Future>::poll(Pin::new(token), cx);
 
-        match this.0.next() {
-            Some(res) => Poll::Ready(res),
+            match poll {
+                Poll::Ready(Ok(_)) => Some((*id, Ok(*id))),
+                Poll::Pending | Poll::Ready(Err(TokenError::Waiting)) => None,
```
lucaato, line 107: We should be wary of the Ready case since futures that return Ready should not be polled again as per rust documentation. From inspection of the library code this should never occur but maybe adding a comment or removing the packet that returned the ready could be a workaround. Logging the weird state could also make sense, you decide

joshuachp: Yes, I also think we should print a warning here if the case occurs. So we can catch it in the logs

joshuachp: After some consideration, I decided to drop the future and return an error to keep the state consistent and don't risk a panic

src/transport/mqtt/retention.rs
```diff
@@ -104,29 +89,42 @@ impl<'a> IntoFuture for &'a mut MqttRetention {
...
+        this.queue();
+
+        let first = this.packets.iter_mut().find_map(|(id, token)| {
+            let poll = <Token<AckOfPub> as Future>::poll(Pin::new(token), cx);
+
+            match poll {
+                Poll::Pending => None,
+                Poll::Ready(Ok(_)) => Some((*id, Ok(*id))),
+                Poll::Ready(Err(TokenError::Waiting)) => {
+                    warn!(%id, "future returned Ready(Waiting), this should not happend and it could lead to errors on the next pool");
```
lucaato, line 109:
```suggestion
                    warn!(%id, "future returned Ready(Waiting), this should not happend and it could lead to errors on the next poll");
```

lucaato (commented):
Just a typo lgtm, i'll approve in advance

lucaato (approved):

harlem88 (approved):

PR #453, opened 2025-05-14
Implement session storage and status book keeping for reconnection

lucaato:


src/transport/mqtt/connection.rs
```diff
@@ -336,6 +361,18 @@ impl State {
             State::Connected(connected) => connected.poll(conn).await,
         };
 
+        if matches!(next, Next::State(State::Connected(..))) {
```
rgallor, line 364:
```suggestion
        if let Next::State(State::Connected(..)) = next {
```

src/transport/mqtt/connection.rs
```diff
@@ -488,26 +531,58 @@ impl Handshake {
...
+        S: PropertyStore + StoreCapabilities,
     {
         let client = conn.client.clone();
+        let client_id: ClientId = client_id.into();
+        let store = store.clone();
 
-        let client_id_cl: ClientId = client_id.into();
+        let handle = if self.session_present && conn.session_sync {
+            info!(
+                "Already synchronized skipping handshake (introspection synchronized: {})",
```
rgallor, line 542: i would simplify in something like 

```suggestion
                "Introspection already synchronized: {}",
```

src/transport/mqtt/connection.rs
```diff
@@ -488,26 +531,58 @@ impl Handshake {
...
+                conn.session_sync
+            );
 
-        let store = store.clone();
-        let handle: JoinHandle<Result<(), InitError>> = tokio::spawn(async move {
-            let client_id = client_id_cl.as_ref();
+            Self::prop_handshake(client, client_id, store, session_data, self.session_present)
+        } else {
+            info!(
+                "The device is not synchronized (session_present: {}, matching introspection: {}), redoing handshake",
```
rgallor, line 549: also here, something like:

```suggestion
                "Perform again handshake to synchronize the device",
```

the actual log could be good in DEBUG mode, but imo it's too long for INFO level

src/transport/mqtt/connection.rs
```diff
@@ -523,9 +598,37 @@ impl Handshake {
...
+        client: AsyncClient,
+        client_id: ClientId,
+        store: StoreWrapper<S>,
+        session_data: SessionData,
+        session_present: bool,
+    ) -> JoinHandle<Result<(), InitError>>
+    where
+        S: PropertyStore,
+    {
+        // FIXME we should probably purge the properties
```
rgallor, line 614: do you want to fix it in this PR?

src/session/mod.rs
```diff
@@ -0,0 +1,195 @@
...
+    pub(crate) fn matches(&self, stored: &[IntrospectionInterface]) -> bool {
+        stored.len() == self.len()
+            && stored.iter().all(|stored_i| {
+                self.get(&stored_i.name).is_some_and(|i| {
+                    i.version_major() == stored_i.version_major
+                        && i.version_minor() == stored_i.version_minor
+                })
+            })
+    }
+}
```
joshuachp, line 134: This should be moved to interfaces.rs

src/store/mod.rs
```diff
@@ -48,11 +49,25 @@ pub trait StoreCapabilities: PropertyStore {
...
+    /// Returns the introspection store if supported.
+    fn get_session(&self) -> Option<&Self::Session>;
 }
 
+/// Un-constructable type for a default capability.
+///
+/// This should be the never type [`!`] in the future.
+/// Useful for types which do not have a capability but must implement [`StoreCapabilities`]
+#[derive(Clone, Copy)]
+pub enum MissingCapability {}
```
joshuachp, line 69: Move here all the implementations for missing capabilities

src/transport/mqtt/connection.rs
```diff
@@ -283,12 +280,36 @@ struct Connection {
...
+    // Wether the stored introspection matches the current one
+    session_sync: bool,
 }
 
 impl Connection {
     fn eventloop_mut(&mut self) -> &mut EventLoop {
         self.eventloop.get_mut()
     }
+
+    async fn update_session_status<S>(&mut self, interfaces: &Interfaces, store: &StoreWrapper<S>)
```
joshuachp, line 292: I would make this function return the session status instead of changing it, so we make sure it's false in the error case and early returns
```suggestion
    #[must_use]
    async fn update_session_status<S>(&mut self, interfaces: &Interfaces, store: &StoreWrapper<S>) -> bool
```

joshuachp (changes requested):

joshuachp (changes requested):

src/transport/mqtt/connection.rs
lucaato: Yes i fixed it

src/transport/mqtt/connection.rs
lucaato: Moved this code in the `wait_connack` function, to make it run only on first connection

src/session/sqlite/statement.rs
```diff
@@ -0,0 +1,112 @@
...
+        })
+    }
+
+    pub(crate) fn clear_introspection(&self) -> Result<(), SqliteError> {
+        wrap_sync_call(|| {
+            let mut statement = self.prepare_cached(include_query!(
+                "queries/session/write/clear_introspection.sql"
+            ))?;
+
+            statement.execute(())?;
```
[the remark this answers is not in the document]
lucaato: I don't think because I haven't opened a transaction since this is a single query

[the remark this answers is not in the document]
rgallor: Make sense 🔝

src/transport/mqtt/connection.rs
```diff
@@ -488,44 +530,101 @@ impl Handshake {
...
+        client: AsyncClient,
+        client_id: ClientId,
+        store: StoreWrapper<S>,
+        session_data: SessionData,
+    ) -> JoinHandle<Result<(), InitError>>
+    where
+        S: PropertyStore,
+    {
+        tokio::spawn(async move {
+            let client_id = client_id.as_ref();
```
harlem88, line 569: When a new full handshake start, maybe you have to set the `session_sync = false`, so you can avoid unexpected behavior.

lucaato: Since we have a shared reference to the `Interfaces` struct new interfaces can't be added or removed so performing resetting the synchronization status seems to be an overly safe thing to do. I'm can't think of any situation where that would be correct an error, but we can do it to avoid possible problems. We can decide this with @joshuachp and @rgallor

src/client/introspection.rs
```diff
@@ -98,7 +98,14 @@ where
 {
     async fn add_interface(&mut self, interface: Interface) -> Result<bool, Error> {
         // Lock for writing for the whole scope, even the checks
-        let mut interfaces = self.state.interfaces.write().await;
+        let _permit = self
+            .state
+            .introspection
+            .acquire()
+            .await
+            .map_err(|_| Error::Disconnected)?;
```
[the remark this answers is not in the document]
lucaato: In theory this error should not occur as is returned only when the semaphore is closed, and we do not call [`close`](https://docs.rs/tokio/1.43.0/tokio/sync/struct.Semaphore.html#method.close). However this could be useful in the future to signal a disconnected device.

src/transport/mqtt/connection.rs
```diff
@@ -969,130 +969,226 @@ impl Next {
...
+                .expect_subscribe::<String>()
+                .once()
+                .in_sequence(seq)
+                .with(
+                    predicate::eq(format!("{}/control/consumer/properties", self.client_id)),
+                    predicate::always(),
+                )
+                .returning(|_topic, _qos| notify_success(SubAck::new(0, Vec::new())));
+
+            for si in &self.server_interfaces {
```
rgallor, line 1039: such an Italian abreviation 😆

rgallor (changes requested):

src/store/mock.rs
```diff
@@ -0,0 +1,128 @@
...
+            interface: &super::PropertyInterface<'a>,
+        ) -> Result<(), StoreError>;
+
+        async fn device_props_with_unset(
+            &self,
+        ) -> Result<Vec<super::OptStoredProp>, StoreError>;
+    }
+
+    impl MockedStoreCapabilities for Store {
+        // fn return_retention(&self) -> bool;
```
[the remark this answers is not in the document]
lucaato: Enabled it since when the mock for the retention gets implemented it should get enabled

src/client/introspection.rs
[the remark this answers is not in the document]
lucaato: We decided it is better to close the semaphore when the connection gets dropped so we implemented `Drop` for the `DeviceConnection`

src/client/introspection.rs
```diff
@@ -126,7 +137,13 @@ where
         I: IntoIterator<Item = Interface> + Send,
     {
         // Lock for writing for the whole scope, even the checks
-        let mut interfaces = self.state.interfaces.write().await;
+        let _permit = self
+            .state
+            .introspection
+            .acquire()
+            .await
+            .map_err(|_| Error::Disconnected)?;
```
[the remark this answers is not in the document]
lucaato: I answered in the other comment.

joshuachp (changes requested):

src/transport/mqtt/connection.rs
joshuachp: Yes I think that would be a good thing, because we need to keep track that the session present was false (at least once).

lucaato: Reset the session sync and cleared the store in the case of a full handshake, i then store a `session_sync = true` only when we join successfully the `full_handshake` task

src/session/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+
+use crate::{error::DynError, interfaces::Interfaces, Interface};
+
+mod sqlite;
+
+/// Interface data associated with the astarte introspection.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct IntrospectionInterface {
+    /// Name of the interface.
+    name: String,
```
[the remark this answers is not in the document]
lucaato: I'm not sure whether this is worth the time and pollution of the types with a reference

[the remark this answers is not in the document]
lucaato: Added a generic parameter that defaults to String

src/interfaces.rs
```diff
@@ -252,6 +253,24 @@ impl Interfaces {
...
+        stored.len() == self.len()
+            && stored.iter().all(|stored_i| {
+                self.get(stored_i.name().as_ref()).is_some_and(|i| {
+                    i.version_major() == stored_i.version_major()
+                        && i.version_minor() == stored_i.version_minor()
+                })
+            })
+    }
+
+    pub(crate) fn len(&self) -> usize {
```
rgallor, line 267: you can impl Deref for Interfaces and avoid redefining len and is_empty hashmap methods

rgallor (changes requested):

src/session/mod.rs
```diff
@@ -0,0 +1,248 @@
...
+use std::future::Future;
+
+use itertools::Itertools;
+
+use crate::{error::DynError, interfaces::Interfaces, Interface};
+
+mod sqlite;
+
+/// Interface data associated with the astarte introspection.
+#[derive(Debug, Clone, PartialOrd, Ord, Hash)]
```
joshuachp, line 31:
```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
```

joshuachp: I don't know if `PartialEq` is possible or would conflict with the manual impl.

But we should derive copy

src/session/mod.rs
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    pub(crate) fn remove_interfaces(err: impl Into<DynError>) -> Self {
+        Self::RemoveInterfaces(err.into())
+    }
+}
+
+/// Trait for persistently storing and managing session-related data.
+pub trait StoredSession: Clone + Send + Sync {
+    /// Clears all `IntrospectionInterface`s from the persistent store.
```
joshuachp, line 106:
```suggestion
    /// Clears all [`IntrospectionInterface`] from the persistent store.
```

src/session/mod.rs
```diff
@@ -0,0 +1,248 @@
...
+    fn from(val: &'a Interfaces) -> Self {
+        val.iter().map(|i| i.into()).collect_vec()
+    }
+}
+
+impl From<&Interfaces> for Vec<IntrospectionInterface> {
+    fn from(val: &Interfaces) -> Self {
+        val.iter().map(|i| i.into()).collect_vec()
+    }
+}
```
joshuachp, line 169: Move these traits below the struct declaration

src/transport/mqtt/connection.rs
```diff
@@ -399,9 +414,12 @@ impl Disconnected {
...
         debug!("created a new transport, reconnecting");
 
-        let eventloop = conn.eventloop_mut();
-        eventloop.clean();
-        Self::set_transport(eventloop, transport);
+        if conn.session_sync {
+            Self::disable_clean_session(conn.eventloop_mut());
+        } else {
+            conn.eventloop_mut().clean();
+        }
```
joshuachp, line 421: Move this when we receive an ConnAck with session present false.

src/transport/mqtt/connection.rs
```diff
@@ -399,9 +414,12 @@ impl Disconnected {
 
         debug!("created a new transport, reconnecting");
 
-        let eventloop = conn.eventloop_mut();
-        eventloop.clean();
-        Self::set_transport(eventloop, transport);
+        if conn.session_sync {
+            Self::disable_clean_session(conn.eventloop_mut());
```
joshuachp, line 418: We can move this when we set conn.session_sync to true in the wait acks

src/transport/mqtt/connection.rs
```diff
@@ -679,44 +772,39 @@ impl WaitAcks {
...
             Ok(Ok(())) => {
                 info!("device connected");
-                Ok(Next::state(Connected))
-            }
-            Ok(Err(InitError::Unset(err))) => {
-                error!(error = %Report::new(&err), "init task failed");
-
-                Err(err)
+                // if the device successfully connects we set the sync status to true
+                conn.session_sync = true;
```
joshuachp, line 801: Here we can disable the clean session

joshuachp (changes requested):

src/transport/mqtt/connection.rs
joshuachp: I think before it was wrong to do it at every reconnection.

lucaato: The only way to perform the clean after connecting is to go back to the Disconnected state again since it resets the connection. The should only be performed while disconnected

src/session/mod.rs
lucaato: ? I don't understand they are already below the struct.

src/transport/mqtt/connection.rs
```diff
@@ -283,12 +296,32 @@ struct Connection {
...
     }
+
+    pub(crate) fn set_session_synced(&mut self, sync: bool) {
+        self.session_synced = sync;
+
+        if sync {
+            Self::set_clean_session(self.eventloop.get_mut(), false);
+        } else {
+            Self::set_clean_session(self.eventloop.get_mut(), true);
+        }
```
joshuachp, line 315:
```suggestion
            // We can not access a field of the mock so we have to add the cfg
            #[cfg(not(test)]
            self.eventloop.get_mut().mqtt_options.set_clean_session(!sync);
```

src/transport/mqtt/connection.rs
```diff
@@ -283,12 +296,32 @@ struct Connection {
...
+    }
+
+    #[cfg(not(test))]
+    fn set_clean_session(eventloop: &mut EventLoop, clean: bool) {
+        eventloop.mqtt_options.set_clean_session(clean);
+    }
+
+    #[cfg(test)]
+    fn set_clean_session(_eventloop: &mut EventLoop, _clean: bool) {}
```
joshuachp, line 324:
```suggestion
```

joshuachp (approved):
Last thing, the rest looks good to me!

Wonderful job 🎊

PR #455, opened 2025-05-16
feat(sqlite): set max db size

### LINKED ISSUE #450: [Config]: Set the maximum number of pages in the database file
Currently the database has no limit on the size of database file. 
Expose in the SDK Configuration a way to configure this limit.

 ref. https://sqlite.org/pragma.html#pragma_max_page_count

rgallor:
Allow setting the sqlite database size by providing either the maximum number of db pages or a size in KB, KiB, MB or MiB. Give also the possibility to set the journal size when building an `SQLiteStore`. Also set default pragma when connecting to an sqlite store.

Closes #450

### DIFF
--- a/src/store/sqlite/mod.rs
+++ b/src/store/sqlite/mod.rs
@@ -18,15 +18,16 @@
 
 //! Provides functionality for instantiating an Astarte sqlite database.
 
-use std::{cell::Cell, fmt::Debug, path::Path, sync::Arc, time::Duration};
+use std::{cell::Cell, fmt::Debug, num::NonZeroU64, path::Path, sync::Arc, time::Duration};
 
 use futures::lock::Mutex;
 use rusqlite::{
     types::{FromSql, FromSqlError},
-    OptionalExtension, ToSql,
+    Connection, OptionalExtension, ToSql,
 };
+use serde::{Deserialize, Serialize};
 use statements::{include_query, ReadConnection, WriteConnection};
-use tracing::{debug, error, trace};
+use tracing::{debug, error, trace, warn};
 
 use super::{
     OptStoredProp, PropertyInterface, PropertyMapping, PropertyStore, StoreCapabilities, StoredProp,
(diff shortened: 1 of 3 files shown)


src/store/sqlite/mod.rs
```diff
@@ -645,4 +772,131 @@ mod tests {
...
+                        vec![0u8, 255],
+                        vec![1u8, 255],
+                        vec![2u8, 255],
+                        vec![3u8, 255],
+                    ],
+                )
+                .unwrap();
+        }
+
+        let err = db.set_max_pages(1).await.unwrap_err();
```
joshuachp, line 836: Store some properties of type BinaryBlob instead of creating a new table.

joshuachp (changes requested):
Add a test that will go over the maximum size of the db

joshuachp (approved):

harlem88 (commented):
I think we should add a way to config the size in a programmatic way, so for example we could add an option to config the size limit on DeviceBuilder:
``` rust
DeviceBuilder::new()
        .store_limit(1MiB)
        .store_dir(&tmp_dir)
```
What do you think? So if we want to add a different limit for example for message-hub we don't have to change it and use a specific version of sdk

rgallor:
> I think we should add a way to config the size in a programmatic way, so for example we could add an option to config the size limit on DeviceBuilder:
> 
> ```rust
> DeviceBuilder::new()
>         .store_limit(1MiB)
>         .store_dir(&tmp_dir)
> ```
> 
> What do you think? So if we want to add a different limit for example for message-hub we don't have to change it and use a specific version of sdk

I considered your idea and i found it also a plausible solution, but i didn't like too much the idea of adding fields to the builder to set the db size and also other parameterrs (such as the journal size). Also, a check must be done in case the store dir is not provided and the other fields are not.
I found it more direct to directly work on the SQLiteStore and pass it through the `DeviceBuilder::store()` method.

harlem88 (commented):
LGTM, add these changes also in the Changelog

src/store/sqlite/mod.rs
```diff
@@ -151,6 +177,56 @@ pub enum ValueError {
...
+impl Size {
+    /// Convert the size to bytes
+    const fn to_bytes(self) -> u64 {
+        match self {
+            Size::Kb(kb) => kb.get().saturating_mul(1000),
+            Size::Mb(mb) => mb.get().saturating_mul(1000 * 1000),
+            Size::GB(gb) => gb.get().saturating_mul(1000 * 1000 * 1000),
+            Size::KiB(kib) => kib.get().saturating_mul(1024),
+            Size::MiB(mib) => mib.get().saturating_mul(1024 * 1024),
+            Size::GiB(gib) => gib.get().saturating_mul(1024 * 1024 * 1024),
```
joshuachp, line 213: Add a test for each unit to check the conversion to_bytes and to_kibi

Ex. `assert_eq!(Size::Kb(1), 1000)`

src/store/sqlite/mod.rs
```diff
@@ -596,6 +738,15 @@ fn deserialize_prop(stored_type: u8, buf: &[u8]) -> Result<AstarteType, ValueErr
...
 }
 
+/// Necessary for rust 1.78 const compatibility
+const fn const_non_zero(v: u64) -> NonZeroU64 {
+    let Some(v) = NonZeroU64::new(v) else {
+        panic!("value cannot be zero");
+    };
+
+    v
+}
```
joshuachp, line 748: Add a test for:

```rust
#[test]
#[should_panic(expected = "value cannot be zero")]
fn const_non_zero_should_panic() {
    const_non_zero(0);
}

File: src/store/sqlite/statements.rs [hunk 4]
```diff
@@ -69,23 +72,46 @@ impl WriteConnection {
...
+        }
+
+        // perform vacuum
+        connection.execute("VACUUM", [])?;
+
+        if db_created {
+            let page_size: u64 = connection.get_pragma("page_size")?;
+            let pages = SQLITE_DEFAULT_DB_MAX_SIZE.calculate_max_page_count(page_size);
+            set_pragma(&connection, "max_page_count", pages)?;
+        }
```
joshuachp, line 99: Add a test that a custom jusrnal_size doesn't change

joshuachp (changes requested):

harlem88 (approved):

PR #459, opened 2025-05-21
refactor(type): create the Float struct to make sure floats are valid

joshuachp:
Since the AstarteType is a public enum it can be constructed with invalid values.


rgallor (approved):

src/types.rs
```diff
@@ -443,6 +482,64 @@ impl ArrayType {
     }
 }
 
+/// A valid Astarte float.
+///
+/// It cannot be Nan, Infinite or subnormal.
+#[derive(Debug, Clone, Copy, PartialEq, PartialOrd)]
+#[repr(transparent)]
+pub struct Float(f64);
```
harlem88, line 490: Since it contains an f64, why don't call directly `Double`? `Double(Float),` it is little strange, I think

harlem88 (changes requested):

src/types.rs
joshuachp: Sure, make sense. I wanted to call it F64 but was it looked wrong so it became Float. But double make sense.

src/transport/grpc/convert.rs
```diff
@@ -114,12 +110,12 @@ fn convert_timestamp(
     let nanos = val
         .nanos
         .try_into()
-        .map_err(|err: TryFromIntError| MessageHubProtoError::DateConversion(err.to_string()))?;
+        .map_err(|_| MessageHubProtoError::Timestamp(val))?;
```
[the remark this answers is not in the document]
joshuachp: The error doesn't have information, https://doc.rust-lang.org/src/core/num/error.rs.html#10

Further more, we provide a more correct error by returning `MessageHubProtoError::Timestamp(val)` since it is actually a conversion of the timestamp and not some integer

lucaato (approved):

PR #462, opened 2025-05-22
feat(types): implement display for AstarteData 

joshuachp:


src/types/display.rs
```diff
@@ -0,0 +1,125 @@
...
+        write!(f, "  ")?;
+        i.fmt(f)?;
+        writeln!(f, ",")?;
+    }
+
+    writeln!(f, "]")
+}
+
+#[cfg(test)]
+mod tests {}
```
rgallor, line 125: remove this i no test has to be inserted

rgallor (changes requested):

src/types/display.rs
```diff
@@ -0,0 +1,204 @@
...
+
+use std::fmt::Display;
+
+use base64::display::Base64Display;
+
+use super::AstarteData;
+
+const BASE64: base64::engine::GeneralPurpose = base64::engine::general_purpose::URL_SAFE;
+
+impl Display for AstarteData {
```
lucaato, line 29: The only thing I'm doubting is if we need a display implementation for Astarte data, this looks more like a debug implementation since there are different types in play and we are also printing out arrays, if you thought that through it's ok either way.

lucaato (approved):

PR #473, opened 2025-06-17
chore(interfaces): warn when an interface name ends in `.json`

### LINKED ISSUE #472: Add a warning when interface name ends in .json
Add a warning when getting an interface from `Interfaces` to check and warn if the interface name ends in `.json`

joshuachp:
Closes #472

### DIFF
--- a/src/interfaces.rs
+++ b/src/interfaces.rs
@@ -28,7 +28,7 @@ use astarte_interfaces::{
     AggregationIndividual, DatastreamIndividual, DatastreamObject, MappingPath, Properties, Schema,
 };
 use itertools::Itertools;
-use tracing::{debug, trace};
+use tracing::{debug, trace, warn};
 
 use crate::error::AggregationError;
 use crate::session::IntrospectionInterface;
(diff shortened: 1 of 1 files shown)


src/interfaces.rs
```diff
@@ -114,6 +114,11 @@ impl Interfaces {
     }
 
     pub(crate) fn get(&self, interface_name: &str) -> Option<&Interface> {
+        #[cfg(debug_assertions)]
+        if interface_name.ends_with(".json") {
+            warn!(interface_name, "the interface name passed ends in .json, this is commonly a BUG when the extension is copied from the interface file and shoulb probably be removed");
```
lucaato, line 119: What about something like this ?
```suggestion
            warn!(interface_name, "the interface name passed ends in \".json\", the input must be an interface name copied from the interface definition \"name\" field");
```

joshuachp: I think the warning is more explicative of what happened if we just ask to remove the file extension

lucaato:
```suggestion
            warn!(interface_name, "the interface name passed ends in .json, this is commonly a BUG when the extension is copied from the interface file and should probably be removed");
```

PR #477, opened 2025-06-25
refactor(mapping): maintain same mappings order when creating an Interface object

rgallor:


joshuachp (changes requested):
Add the following test:

```rust
    #[test]
    fn should_maintain_mapping_order_serde() {
        let original = DatastreamObject::from_str(E2E_DEVICE_AGGREGATE).unwrap();

        let serialized = serde_json::to_string(&original).unwrap();
        let deserialized = DatastreamObject::from_str(&serialized).unwrap();

        assert_eq!(deserialized, original);
    }
```

joshuachp (changes requested):

joshuachp (approved):

PR #482, opened 2025-07-09
ci: run examples

lucaato:


joshuachp (changes requested):
We should standardize all the examples, by using clap and the env feature. Because this will make them more consistent and easier to run and script.

We should add an optional argument for the number of packets to send/receive or a timeout for the example, so we don't need to kill the process, and it will itself exit.

PR #483, opened 2025-07-11
fix(sqlite): Add serde content attribute

lucaato:
To correctly serialize/deserialize the Size struct.


src/store/sqlite/mod.rs
```diff
@@ -995,4 +995,15 @@ mod tests {
     fn const_non_zero_should_panic() {
         const_non_zero(0);
     }
+
+    #[test]
+    fn should_serialize_deserialize_size() {
+        let size = Size::Gb(2.try_into().unwrap());
+
+        let out = serde_json::to_string(&size).unwrap();
```
joshuachp, line 1003: I would assert_eq to the actual string to make sure the serialization deserialization is consistent

joshuachp (changes requested):

joshuachp (approved):

PR #488, opened 2025-07-17
fix(sqlite): Set autocheckpoint size to limit the size of the WAL

lucaato:
The WAL size is not limited by the `journal_size_limit` pragma.
That value sets the size at which the journal will be truncated.
To correctly limit the size of a WAL in use we have to set the `wal_autocheckpoint` pragma to a value lower than the size limit.

### DIFF
--- a/src/store/sqlite/options.rs
+++ b/src/store/sqlite/options.rs
@@ -28,21 +28,33 @@ use super::{
     SQLITE_DEFAULT_DB_MAX_SIZE, SQLITE_JOURNAL_SIZE_LIMIT, SQLITE_WAL_AUTOCHECKPOINT,
 };
 
+/// Choices of limit of the size of the sqlite database
+#[derive(Clone, Debug, Copy)]
+pub enum SizeLimit {
+    /// Set a size limit based on the number of pages
+    Pages(u32),
+    /// Set a size limit based on the actual size of the db file
+    ///
+    /// This value will be approxymated if it's not a multiple of the databse page size
+    Size(Size),
+}
+
 /// Sqlite options that can be set externally to tweak the behaviour of sqlite.
+#[derive(Clone, Debug)]
 pub struct SqliteStoreOptions {
     // Maximum number of pages in the sqlite db file
-    db_size_limit: Size,
+    pub(crate) db_size_limit: SizeLimit,
     /// Since we use WAL mode the size of our journal controls the size of the WAL
     /// Without the autocheckpoint set the journal still grows until an autocheckpoint is performed.
     /// The size limit only applies when the WAL journal is truncated. We set both options to
     /// correctly limit the size of the WAL file.
-    journal_size_limit: Size,
+    pub(crate) journal_size_limit: Size,
 }
 
 impl Default for SqliteStoreOptions {
     fn default() -> Self {
         Self {
-            db_size_limit: SQLITE_DEFAULT_DB_MAX_SIZE,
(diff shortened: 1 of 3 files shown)


joshuachp:
🤖

joshuachp (changes requested):
I would move add the SqlitePragma struct to the constructor of the connection, and move all the pragmas invocation to this struct. They don't need all to be configurable for now, but we should apply the same pragmas to both the reader and writer connection.

Because this will make sure all the option are the same for every connection and simplify the code, having all the pragmas in just one place

We should move the SqliteOption and SqlitePragmas to their own file.

src/store/sqlite/options.rs
```diff
@@ -0,0 +1,168 @@
...
+    get_pragma, set_pragma, Size, SqliteError, SQLITE_BUSY_TIMEOUT, SQLITE_CACHE_SIZE,
+    SQLITE_DEFAULT_DB_MAX_SIZE, SQLITE_JOURNAL_SIZE_LIMIT, SQLITE_WAL_AUTOCHECKPOINT,
+};
+
+/// Choices of limit of the size of the sqlite database
+#[derive(Clone, Debug, Copy)]
+pub enum SizeLimit {
+    /// Set a size limit based on the number of pages
+    Pages(u32),
+    /// Set a size limit based on the actual size of the db file
```
joshuachp, line 36:
```suggestion
    /// Set a size limit based on the actual size of the db file
    ///
    /// This value will be approxymated if it's not a multiple of the databse page size
```

src/store/sqlite/options.rs
```diff
@@ -0,0 +1,168 @@
...
+        set_pragma(connection, "journal_size_limit", self.journal_size_limit)?;
+        set_pragma(connection, "wal_autocheckpoint", self.wal_autocheckpoint)?;
+        set_pragma(connection, "foreign_keys", true)?;
+        set_pragma(connection, "busy_timeout", SQLITE_BUSY_TIMEOUT)?;
+        set_pragma(connection, "synchronous", "NORMAL")?;
+        // Reduces the size of the database
+        set_pragma(connection, "auto_vacuum", "INCREMENTAL")?;
+        set_pragma(connection, "temp_store", "MEMORY")?;
+        set_pragma(connection, "cache_size", SQLITE_CACHE_SIZE)?;
+        set_pragma(connection, "journal_mode", "WAL")?;
```
joshuachp, line 164:
```suggestion
    pub(crate) fn apply_pragmas(&self, connection: &Connection) -> Result<(), SqliteError> {
        set_pragma(connection, "journal_mode", "WAL")?;
        set_pragma(connection, "max_page_count", self.max_page_count)?;
        set_pragma(connection, "journal_size_limit", self.journal_size_limit)?;
        set_pragma(connection, "wal_autocheckpoint", self.wal_autocheckpoint)?;
        set_pragma(connection, "foreign_keys", true)?;
        set_pragma(connection, "busy_timeout", SQLITE_BUSY_TIMEOUT)?;
        set_pragma(connection, "synchronous", "NORMAL")?;
        // Reduces the size of the database
        set_pragma(connection, "auto_vacuum", "INCREMENTAL")?;
        set_pragma(connection, "temp_store", "MEMORY")?;
        set_pragma(connection, "cache_size", SQLITE_CACHE_SIZE)?;

```

joshuachp: I would put the wal first

joshuachp (changes requested):

PR #495, opened 2025-07-28
fix(retention): Fix object send not correctly handling reliable messages

lucaato:


src/transport/mqtt/mod.rs
```diff
@@ -1301,4 +1295,251 @@ pub(crate) mod test {
         mock_url.assert_async().await;
         mock_cert.assert_async().await;
     }
+
+    #[tokio::test]
+    async fn should_not_enque_individual() {
```
joshuachp, line 1300:
```suggestion
    async fn should_not_enque_sent_individual() {
```

src/transport/mqtt/mod.rs
```diff
@@ -1301,4 +1295,251 @@ pub(crate) mod test {
...
+                ("longinteger".to_string(), AstarteType::LongInteger(10)),
+                ("boolean".to_string(), AstarteType::Boolean(false)),
+            ]
+            .into_iter()
+            .collect())
+        }
+    }
+
+    #[tokio::test]
+    async fn should_not_enque_object() {
```
joshuachp, line 1395:
```suggestion
    async fn should_not_enque_sent_object() {
```

joshuachp (changes requested):

joshuachp (approved):

src/transport/mqtt/mod.rs
```diff
@@ -225,18 +225,35 @@ impl<S> MqttClient<S> {
...
-                if let Some(retention) = self.store.get_retention() {
-                    retention.mark_as_sent(id).await?;
+            Reliability::Guaranteed | Reliability::Unique => {
+                match id {
+                    RetentionId::Volatile(id) => {
+                        self.volatile.mark_sent(&id, true).await;
+                    }
+                    RetentionId::Stored(id) => {
+                        if let Some(retention) = self.store.get_retention() {
+                            retention.mark_as_sent(&id).await?;
```
harlem88, line 249: Why didn't we call all methods like `mark_sent`? have they a different behavior?
```suggestion
                            retention.mark_sent(&id).await?;
```

harlem88 (changes requested):

harlem88 (approved):

PR #508, opened 2025-09-17
fix(bson): serialize timestamp as datetime

joshuachp:


README.md
```diff
@@ -1,5 +1,20 @@
 <!--
+This file is part of Astarte.
+
 Copyright 2021,2022 SECO Mind Srl
+Copyright 2025 SECO Mind Srl
```
harlem88, line 5: Maybe here Do we have a double date?

joshuachp: fixed

PR #519, opened 2025-09-30
Feature/validate device certificate following error

lucaato:


joshuachp (changes requested):

src/transport/mqtt/config/tls.rs
```diff
@@ -74,57 +71,96 @@ impl ClientAuth {
...
     pub(crate) fn try_from_pem_cert(
-        certs: String,
+        pem: String,
         private_key: PrivatePkcs8KeyDer<'static>,
-    ) -> Result<Self, io::Error> {
-        let certs = rustls_pemfile::certs(&mut certs.as_bytes()).try_collect()?;
+    ) -> Result<Option<Self>, io::Error> {
+        let Some(cert) = rustls_pemfile::certs(&mut pem.as_bytes()).next() else {
+            return Ok(None);
+        };
```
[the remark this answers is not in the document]
joshuachp: or iter tools provides exactly_once

src/transport/mqtt/config/tls.rs
```diff
@@ -74,57 +71,96 @@ impl ClientAuth {
...
+            Some(Self {
+                der: cert,
+                private_key,
+                pem,
+            })
+        })
     }
 
     /// Blocking function to read the certificate and the key
     fn read_cert_and_key(
```
joshuachp:
```suggestion
    async fn read_cert_and_key(
```

src/transport/mqtt/config/tls.rs
```diff
@@ -74,57 +71,96 @@ impl ClientAuth {
...
     ) -> Result<Option<Self>, io::Error> {
-        let c_f = File::open(certificates)?;
-
-        let mut c_r = BufReader::new(c_f);
-        let certs: Vec<CertificateDer<'static>> = rustls_pemfile::certs(&mut c_r).try_collect()?;
-
-        if certs.is_empty() {
-            debug!("no certificate found");
-
+        let pem = fs::read_to_string(certificates.path())?;
```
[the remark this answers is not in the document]
joshuachp: Before this function used a BufReader, now you changed it to a fs::read_to_string, so you must use the Tokyo version to not block.

src/transport/mqtt/config/tls.rs
joshuachp: I would verify the certificate with the client id and date when reading it from disk instead of having it as a separte step

src/transport/mqtt/config/tls.rs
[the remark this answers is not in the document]
joshuachp: Also this should never return an option now, since it parses the Pem file from Astarte

src/transport/mqtt/config/tls.rs
```diff
@@ -45,7 +41,8 @@ pub(crate) fn is_env_ignore_ssl() -> bool {
 }
 
 pub(crate) struct ClientAuth {
-    certs: Vec<CertificateDer<'static>>,
+    pem: String,
+    der: CertificateDer<'static>,
```
joshuachp:
```suggestion
    der: Vec<CertificateDer<'static>>,
```

I was wrong in my previous suggestion, this should be an array, with certificates for multiple origins

joshuachp (commented):
I would store the full certs vec and validate each (contrary on what said before)

src/transport/mqtt/config/tls.rs
lucaato: Did this ever happen and do we expect it to happen, have you asked the back-end team ? We can't base changes off intuition (i probably shouldn't have changed it the first time around) but I want to know if certificate for multiple origins are allowed first.

src/transport/mqtt/config/tls.rs
lucaato: We have no way of knowing whether the certificate is actually valid since it has also to be verified by the server too and the whole pr started as a check in the event of an error. So I think it's more of a matter of preference, I would leave it here unless there's something that actually makes the current implementation not usable.

src/transport/mqtt/config/transport.rs
```diff
@@ -166,8 +168,11 @@ impl TransportProvider {
             .await
         }
 
-        ClientAuth::try_from_pem_cert(certificate, bundle.private_key)
-            .map_err(PairingError::InvalidCredentials)
+        match ClientAuth::try_from_pem_cert(certificate, bundle.private_key) {
+            Ok(Some(auth)) => Ok(auth),
+            Ok(None) => Err(PairingError::MissingCredentials),
+            Err(e) => Err(PairingError::InvalidCredentials(e)),
+        }
```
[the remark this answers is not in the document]
lucaato: As answered above, I'd prefer not to change it, since it better represents the possible output of the function.

src/transport/mqtt/config/tls.rs
joshuachp: The only single place this is called, the null is returned as an error, so it is simpler to move it here.

This is called only with the Astarte PEM, so i would rename it to try_from_astarte_cert

src/transport/mqtt/config/tls.rs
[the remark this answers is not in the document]
joshuachp: Yes, I also ment to remove the souraundind spawn to not have to guess when using this function whether it needs a spawn_blocking

src/transport/mqtt/config/tls.rs
joshuachp: I only mean the local checks and not the API calls, to prevent easy to check mistakes and report them

src/transport/mqtt/config/tls.rs
lucaato: Upon further consideration and irl discussion i think I will leave this function as is since i would have tho change the error type, and that would be the same as handling it outside. This also leaves this function more reusable.

src/transport/mqtt/config/tls.rs
lucaato: Verified in the transport in retrieve credentials

src/transport/mqtt/config/transport.rs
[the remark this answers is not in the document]
lucaato: I'm keeping my position about this change as above

src/transport/mqtt/config/tls.rs
joshuachp: Let's leave it like this if you don't want to change it

lucaato: You're right I'll reuse the function below

src/transport/mqtt/config/tls.rs
lucaato: It is now verified in the transport since we need the client id

src/transport/mqtt/config/tls.rs
```diff
@@ -74,57 +69,87 @@ impl ClientAuth {
     }
 
     pub(crate) fn try_from_pem_cert(
-        certs: String,
+        pem: String,
         private_key: PrivatePkcs8KeyDer<'static>,
-    ) -> Result<Self, io::Error> {
-        let certs = rustls_pemfile::certs(&mut certs.as_bytes()).try_collect()?;
+    ) -> Result<Option<Self>, io::Error> {
+        let Some(cert) = rustls_pemfile::certs(&mut pem.as_bytes()).next() else {
```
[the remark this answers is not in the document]
joshuachp: from itertools

PR #604, opened 2026-05-05
refactor(pairing): improve logging and error handling

joshuachp:
Show traces for the configured URL and show the response for the API calls.


src/transport/mqtt/pairing.rs
```diff
@@ -225,6 +228,47 @@ impl<'a> ApiClient<'a> {
...
+                return;
+            }
+        };
+
+        match serde_json::from_slice::<serde_json::Value>(&body.to_bytes()) {
+            Ok(json) => {
+                error!(%json, "HTTP error response");
+            }
+            Err(error) => {
+                error!(%error,  "HTTP error JSON response received, but was not a valid");
```
lucaato:
```suggestion
                error!(%error, "HTTP error JSON response received, but was not a valid");
```

src/transport/mqtt/pairing.rs
```diff
@@ -225,6 +228,47 @@ impl<'a> ApiClient<'a> {
...
+                trace!("invalid content-type header");
+
+                return false;
+            };
+
+            APPLICATION_JSON == value
+        });
+
+        if !is_json {
+            error!("HTTP error response received, but does't have a JSON content-type");
```
lucaato, line 246: would it make sense to log part of the response (with a limit) even when not json ?

joshuachp: Astarte should only respond with json, other stuff we should consider garbage and the status code should be sufficient.

For example, if we receive a 502 for Bad Gateway the body will be the HTML for the nginx/haproxy ingress; which I don't think we should print and has the same value as the status code.

lucaato: If you're sure about it

lucaato (approved):

PR #605, opened 2026-05-06
refactor(connection): pair ascyncronously while connecting 

joshuachp:


src/transport/mqtt/config/transport.rs
```diff
@@ -54,35 +48,127 @@ pub(crate) struct TransportProvider {
...
+
+        let auth = self.read_credentials(client_id).await;
+
+        let Some(auth) = auth else {
+            notify_security_event(SecurityEvent::AlarmCertificateUnavailable);
+
+            return Ok(None);
+        };
+
+        info!("existing certificate found");
```
harlem88:
```suggestion
        debug!("existing certificate found");
```

harlem88 (approved):

src/transport/mqtt/connection/context.rs
```diff
@@ -0,0 +1,105 @@
...
+use crate::interfaces::Interfaces;
+use crate::state::SharedState;
+use crate::store::wrapper::StoreWrapper;
+use crate::transport::mqtt::ClientSender;
+use crate::transport::mqtt::client::{AsyncClient, EventLoop};
+use crate::transport::mqtt::config::transport::TransportProvider;
+
+/// Context for the connection
+#[derive(Debug)]
+pub struct Ctx<'a, S> {
```
lucaato: I'd call this `ConnectionCtx` or something different from `Ctx`

src/transport/mqtt/connection/disconnected.rs
```diff
@@ -0,0 +1,146 @@
...
+                (
+                    Self::create_connection(conn, ctx, cfg, &api, transport).await?,
+                    expiry,
+                )
+            }
+        };
+
+        // NOTE: this will be updated even if returned None, since otherwise we would keep the
+        //       previous expiry
+        *ctx.state.cert_expiry.write().await = expiry;
```
lucaato, line 116:
```suggestion
        let client_auth = if let Some(client_auth) = client_auth {
            debug!("auth present and valid");

            client_auth
        } else {
            debug!("auth invalid");

            ctx.provider
                .create_credentials(&api, cfg.client_id.as_ref())
                .await?
        };

        let expiry = client_auth.validity_not_after();

        let connection = match &mut self.connection {
            Some(connection) => {
                debug!("connection present and valid");

                let transport = ctx.provider.config_transport(client_auth)?;
                connection.set_transport(transport);

                connection
            }
            conn @ None => {
                debug!("connection missing");

                let transport = ctx.provider.config_transport(client_auth)?;

                Self::create_connection(conn, ctx, cfg, &api, transport).await?
            }
        };

        // NOTE: this will be updated even if returned None, since otherwise we would keep the
        //       previous expiry
        *ctx.state.cert_expiry.write().await = expiry;
```

lucaato: consider doing this since it reduces the amount of code
_the only downside I see is that it's resetting the transport when both the connection and auth are valid_

src/pairing/fdo/mod.rs
```diff
@@ -29,105 +30,169 @@ use astarte_device_fdo::srv_info::{AstarteMod, AstarteModBuilder};
...
-// TODO: most of this should be done inside the builder
-impl<'a, C> FdoConfig<'a, C> {
-    /// Returns the builder for the FDO config.
-    pub fn build(model_no: &'a str, serial_no: &'a str) -> FdoConfigBuilder<'a, C, AddStorageDir> {
-        FdoConfigBuilder::new(model_no, serial_no)
-    }
-
-    /// Register the device and configures the MQTT connection
+impl<'a, C> FdoDi<'a, C> {
+    /// Initializes a device to be paired to be then bind to a cloud.
```
lucaato:
```suggestion
    /// Initializes a device to be paired to be then bound to a cloud.
```

lucaato:
like the change gg :top:

src/transport/mqtt/connection/context.rs
joshuachp: ConnCtx

src/transport/mqtt/connection/disconnected.rs
joshuachp: We don't want to recreate the transport in that case. I prefer to be explicit with all the conditions here. The number of branches is the same in both codes.

lucaato: yes but it's way shorter

lucaato (approved):

harlem88 (approved):

PR #614, opened 2026-05-14
feat(object): add required field to interface 

joshuachp:
Add the support to the `required` field that marks a mapping as required.


src/validate/mod.rs
```diff
@@ -209,6 +195,58 @@ impl ValidatedObject {
...
+                    return Err(UserValidationError::ObjectInvalidMapping {
+                        interface: interface.interface_name().to_string(),
+                        path: path.to_string(),
+                        key: key.to_string(),
+                    });
+                }
+                (None, None) => {
+                    // unreachable
+                }
+            }
```
lucaato: Since we have an unreachable `(None, None)` should we preferer a custom enum returned by the iterator ? Or always return `(Option<entry>, mapping)` the mapping not being wrapped in an enum.
Maybe we could also handle the error _before_ the match statement:
```rust
        Iter::new(data.iter(), interface.iter_mappings())
            .map(|(entry, mapping)| {
                if let Some(mapping) = mapping {
                    Ok((entry, mapping))
                } else {
                    Err(UserValidationError::ObjectInvalidMapping {
                        interface: interface.interface_name().to_string(),
                        path: path.to_string(),
                        key: entry.map(|(k, _)| k.to_string()).unwrap_or_default(),
                    })
                }
            })
            .try_for_each(|item| {
                match item? {
                    (Some((key, value)), mapping) => {
                        if !value.eq_mapping_type(mapping.mapping_type()) {
                            return Err(UserValidationError::MappingType {
                                interface: interface.interface_name().to_string(),
                                path: format!("{path}/{key}"),
                                expected: mapping.mapping_type().to_string(),
                                got: value.display_type().to_string(),
                            });
                        }

                        trace!("valid object field {path} {}", value.display_type());
                    }
                    (None, mapping) => {
                        if mapping.required() {
                            return Err(UserValidationError::ObjectMappingRequired {
                                interface: interface.interface_name().to_string(),
                                endpoint: mapping.endpoint().to_string(),
                            });
                        }
                    }
                }

                Ok(())
            })?;
```

joshuachp: Done in the try_for_each.

harlem88 (approved):

src/validate/mod.rs
```diff
@@ -209,6 +195,58 @@ impl ValidatedObject {
             timestamp,
         })
     }
+
+    fn check_mappings(
```
lucaato, line 203: nit: add a comment where you state that data is required to be sorted or sort it in here

src/transport/mqtt/payload.rs
```diff
@@ -363,7 +363,7 @@ mod test {
         let buf = serialize_object(&validated.data, validated.timestamp).unwrap();
 
         let expected = base64::prelude::BASE64_STANDARD
-            .decode("ZQIAAAN2AFICAAABZG91YmxlX2VuZHBvaW50AAAAAAAAABJAEGludGVnZXJfZW5kcG9pbnQA/P///whib29sZWFuX2VuZHBvaW50AAESbG9uZ2ludGVnZXJfZW5kcG9pbnQA7lKbmgoAAAACc3RyaW5nX2VuZHBvaW50AAYAAABoZWxsbwAFYmluYXJ5YmxvYl9lbmRwb2ludAAFAAAAAGhlbGxvCWRhdGV0aW1lX2VuZHBvaW50AEA7YPN6AQAABGRvdWJsZWFycmF5X2VuZHBvaW50ADEAAAABMAAzMzMzMzPzPwExADMzMzMzMwtAATIAZmZmZmZmFkABMwAzMzMzMzMfQAAEaW50ZWdlcmFycmF5X2VuZHBvaW50ACEAAAAQMAABAAAAEDEAAwAAABAyAAUAAAAQMwAHAAAAAARib29sZWFuYXJyYXlfZW5kcG9pbnQAFQAAAAgwAAEIMQAACDIAAQgzAAEABGxvbmdpbnRlZ2VyYXJyYXlfZW5kcG9pbnQAJgAAABIwAO5Sm5oKAAAAEjEA71KbmgoAAAASMgDwUpuaCgAAAAAEc3RyaW5nYXJyYXlfZW5kcG9pbnQAHwAAAAIwAAYAAABoZWxsbwACMQAGAAAAd29ybGQAAARiaW5hcnlibG9iYXJyYXlfZW5kcG9pbnQAHwAAAAUwAAUAAAAAaGVsbG8FMQAFAAAAAHdvcmxkAARkYXRldGltZWFycmF5X2VuZHBvaW50ACYAAAAJMABAO2DzegEAAAkxACg/YPN6AQAACTIAEENg83oBAAAAAAl0ACoAAAAAAAAAAA==")
```
lucaato, line 366: can we use insta for this or is it too much trouble to introduce it now ?

joshuachp: Done

lucaato (approved):

harlem88 (approved):

PR #618, opened 2026-05-19
feat(pairing): add method to check whether the device is paired

joshuachp:
Add a method to the connection to check at runtime whether the device is
binded to a cloud. If a device is bindend the device id, realm and
pairing url cannot be reconfigured. This methods allows user to check
for this case externally.

### DIFF
--- a/src/pairing/api/mod.rs
+++ b/src/pairing/api/mod.rs
@@ -23,7 +23,7 @@ use std::path::{Path, PathBuf};
 
 use reqwest::StatusCode;
 use tokio::fs;
-use tracing::{debug, info};
+use tracing::{debug, error, info, instrument};
 
 use self::client::{ApiClient, ClientArgs};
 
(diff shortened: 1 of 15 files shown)


src/pairing/api/mod.rs
```diff
@@ -232,4 +232,27 @@ impl Pairing for PairingApi {
             secret,
         })
     }
+
+    #[instrument(skip(self), ret)]
+    async fn is_paired(&self, store_dir: Option<&Path>) -> std::io::Result<bool> {
```
lucaato, line 237:
```suggestion
    async fn is_paired(&self, credential_file: Option<&Path>) -> std::io::Result<bool> {
```

src/pairing/fdo/mod.rs
```diff
@@ -263,6 +263,11 @@ where
     {
         self.register(ctx).await
     }
+
+    async fn is_paired(&self, _: Option<&Path>) -> Result<bool, std::io::Error> {
+        // TODO: the FDO is already bided and has an ownership voucher
```
lucaato: typo bound

src/pairing/api/mod.rs
joshuachp: That was a bug, I needed to join the path with the file name

src/pairing/fdo/mod.rs
joshuachp: I changed the comment

harlem88 (approved):

src/pairing/api/mod.rs
lucaato: we got him

src/state.rs
```diff
@@ -98,6 +113,21 @@ impl ClientState {
...
         *self.0.cert_expiry.read().await
     }
+
+    pub(crate) fn is_device_paired(&self) -> bool {
+        let value = DeviceStatus::from(self.0.device_status.load(Ordering::Acquire));
+
+        debug_assert_ne!(
+            value,
+            DeviceStatus::Unknown,
+            "the unknown status should be set in the builder"
```
lucaato:
```suggestion
            "the unknown status should be set only in the builder"
```

lucaato (approved):

PR #623, opened 2026-05-27
feat(tls): add crate to configure tls on the device

joshuachp:
Add a new crate to configure TLS on both the message-hub and edgehog-device-runtime.

### DIFF
--- a/astarte-device-tls/src/lib.rs
+++ b/astarte-device-tls/src/lib.rs
@@ -1,0 +1,176 @@
+// This file is part of Astarte.
+//
+// Copyright 2026 SECO Mind Srl
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+use std::sync::Arc;
+
+use rustls::client::WantsClientCert;
+use rustls::crypto::CryptoProvider;
+use rustls::{ClientConfig, ConfigBuilder, Error, RootCertStore};
+use rustls_native_certs::load_native_certs;
+use tracing::instrument;
+use tracing::{error, info};
+
+// re-export
+pub use rustls;
+
+pub mod insecure;
+
+/// Returns a configured TLS client.
+///
+/// # Errors
+///
+/// If we cannot configure the verifier or read the root CAs.
+#[instrument]
+pub fn config() -> Result<ClientConfig, Error> {
+    builder().map(ConfigBuilder::<ClientConfig, WantsClientCert>::with_no_client_auth)
+}
+
+/// Returns a configured builder.
+///
+/// This builder can be used for client authentication.
+///
+/// # Errors
+///
+/// If we cannot configure the verifier or read the root CAs.
+#[instrument]
+pub fn builder() -> Result<ConfigBuilder<ClientConfig, WantsClientCert>, Error> {
+    // This step is to ensure a CryptoProvider is configured also in the tests.
+    let provider = CryptoProvider::get_default().map_or_else(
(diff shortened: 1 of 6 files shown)


joshuachp:
This PR moves edgehog-tls to de sdk repo, to share the code with the msg-hub

astarte-device-tls/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    cfg_if::cfg_if! {
+        if #[cfg(feature = "platform-verifier")] {
+            platform_verifier(provider)
+        } else if #[cfg(feature = "webpki-roots")]{
+            webpki(provider)
+        } else {
+            _default(provider)
+        }
+    }
```
lucaato, line 66:
```suggestion
cfg_if::cfg_if! {
 37         if #[cfg(feature = "platform-verifier")] {
 38             use rustls_platform_verifier::BuilderVerifierExt;
 39
 40             let config = ClientConfig::builder_with_provider(provider)
 41                 .with_safe_default_protocol_versions()?
 42                 .with_platform_verifier()?;
 44
 45             info!("tls client configured with platform verifier");
 46
 47             Ok(config)
 48         } else if #[cfg(feature = "webpki-roots")]{
 49             let config = ClientConfig::builder_with_provider(provider)
 50                 .with_safe_default_protocol_versions()?
 51                 .with_root_certificates(Arc::clone(get_root_certificates_webpki()));
 53
 54             info!("tls client configured with webpki roots");
 55
 56             Ok(config)
 57         } else {
 58             let config = ClientConfig::builder_with_provider(provider)
 59                 .with_safe_default_protocol_versions()?
 60                 .with_root_certificates(Arc::clone(get_root_certificates_default()));
 62
 63             info!("tls client configured with native roots");
 64
 65             Ok(config)
 66         }
 67     }
```
i'd replace the functions with specific ones to get the root certificates since the rest of the code can fit in a single function

astarte-device-tls/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+}
+
+#[cfg(feature = "platform-verifier")]
```
lucaato, line 69:
```suggestion
#[cfg(feature = "webpki-roots")]
pub fn get_root_certificates_webpki() -> &Arc<RootCertStore> {
    static ROOTS: std::sync::OnceLock<Arc<RootCertStore>> = std::sync::OnceLock::new();

    ROOTS.get_or_init(|| {
        let roots = rustls::RootCertStore {
            roots: webpki_roots::TLS_SERVER_ROOTS.to_vec(),
        };

        Arc::new(roots)
    })
}

#[cfg(not(any(feature = "webpki-roots", feature = "platform-verifier")))]
pub fn get_root_certificates_default() -> &Arc<RootCertStore> {
    static ROOTS: std::sync::OnceLock<Arc<RootCertStore>> = std::sync::OnceLock::new();

    ROOTS.get_or_init(|| {
        let res = rustls_native_certs::load_native_certs();

        for error in res.errors {
            error!(%error, "couldn't load native certificate");
        }

        let mut roots = rustls::RootCertStore::empty();

        let (added, ignored) = roots.add_parsable_certificates(res.certs);

        info!(added, ignored, "native certificates loaded");

        Arc::new(roots)
    })
}
```

astarte-device-tls/src/lib.rs
lucaato: see also the other comment

lucaato (commented):
extracting the functions to get the root certificates could make it easier when integrating with the mqtt astarte client

joshuachp:
My plan is to change the MQTT code or builder to accept a tls ConfigBuilder, which is clonable and we can configure with the client cert, and has already the root certs configured.

I've added a public `builder` functions that can be called by the device runtime and msg-hub and passed to the sdk's builder. This way we don't need to store the root certs and we have already a client configured by passing no auth or the mqtt client cert.

joshuachp:
We could also pass to the SDK a [`ClientConfig`](https://docs.rs/rustls/latest/rustls/client/struct.ClientConfig.html) and building a [`SingleCertAndKey`](https://docs.rs/rustls/latest/rustls/sign/struct.SingleCertAndKey.html) with the MQTT certificate and assigning it to the [`ClientConfig::client_auth_cert_resolver`](https://docs.rs/rustls/latest/rustls/client/struct.ClientConfig.html#structfield.client_auth_cert_resolver) field.


I'm looking into the rustls issue if there is any problem with this method of updating the client auth, but I think is possible since it's a ClientConfig without TLS session information.

lucaato (approved):

harlem88 (approved):

PR #646, opened 2026-06-29
feat(error): introduce an error crate for astarte

joshuachp:
Create a device `astarte-device-error` create to use as a generic error crate.

This improve the thiserror method we were using by having a fixed size
struct that doesn't explode when adding more variants. The current
implementation would have a clippy warning for an error variant of 200+
bytes.

The new implementation is based on an enum ErrorKind without additional
data, that will be used as "error code". We have also a field for static
erorr context, and all the additional dynamic fields like the source and
message are allocated to have a fix size.

This PR also reduces the overall number of errors in the sdk, by
generalizing and reusing the main ones in `src/error.rs`.


src/pairing/mod.rs
```diff
@@ -69,21 +70,27 @@ impl PairingConfig {
...
         let conn_timeout = timeout.as_secs();
         if keep_alive >= conn_timeout {
-            return Err(PairingError::Config(format!(
-                "Keep alive ({keep_alive}s) should be less than the connection timeout ({conn_timeout}s)"
+            return Err(Error::with(
+                PairingApiError::InvalidArgument,
+                "keep alive should be lessa than the connection timeout",
+            )
+            .set_message(format!(
+                "got keep alive ({keep_alive}s) and  connection timeout {conn_timeout}s)"
```
harlem88: nit
```suggestion
                "got keep alive ({keep_alive}s) and connection timeout {conn_timeout}s)"
```

src/error.rs
```diff
@@ -18,198 +18,199 @@
...
+//     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+//         match self {
+//             Self::Connection => write!(f, "connection error"),
+//             Self::Interface(error) => write!(f, "interface error {error}"),
+//             Self::MissingTimestamp => write!(f, "message received with missing explicit timestamp"),
+//             Self::Unset => write!(f, "unset received on property without allow_unset"),
+//             Self::Disconnected => write!(f, "disconnected from astarte"),
+//         }
+//     }
+// }
```
harlem88: remove the code below, if it will no longer be used.
```suggestion
```

harlem88 (changes requested):

harlem88:
Resolve the errors on CI

harlem88 (approved):

astarte-device-error/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+/// - An error kind
+/// - A static context message
+/// - A dynamic display message
+/// - A generic error source
+#[derive(Debug)]
+#[must_use]
+pub struct Error<K> {
+    kind: K,
+    ctx: Option<&'static str>,
+    message: Option<Box<dyn DebugDisplay>>,
```
lucaato: I kind of get why we have two fields but this feels like it could be one. Why don't we provide only the context or include the message in the context inside the display iplementation ? 
Also the name feel inverted regarding their purpose, this is a nit but we should have
```suggestion
    message: Option<&'static str>,
    ctx: Option<Box<dyn DebugDisplay>>,
```

src/event.rs
```diff
@@ -41,124 +40,32 @@ pub struct DeviceEvent {
 }
 
 /// Conversion error from an [`DeviceEvent].
```
lucaato:
```suggestion
/// Conversion error from an [`DeviceEvent`].
```

astarte-device-sdk-derive/src/lib.rs
```diff
@@ -76,20 +78,26 @@ impl ObjectDerive {
...
-                object.insert(#field_n.to_string(), v);
-            }
-        }).collect::<Vec<_>>();
+                errors.handle_in(||
+                    field.field_name(self.rename_all).ok_or_else(||
+                        darling::Error::custom( "missing struct fields")
+                            .with_span(&self.ident)
+                    ))
+                    .map(|(field_i, field_n)| {
+                        if field.failable {
```
lucaato: the other sounds wierd to me
```suggestion
                        if field.fallible {
```

src/error.rs
```diff
@@ -18,198 +18,141 @@
...
-    exp: InterfaceType,
-    /// Actual interface type.
-    got: InterfaceType,
+/// Error for the Astarte Interfaces
+///
+/// Returned when an error is caused by an Interface.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum InterfaceError {
+    /// Invalid interface
```
lucaato, line 87: should we consider using displaydoc ?

src/client/mod.rs
```diff
@@ -331,23 +258,25 @@ pub trait Client: Clone {
...
     /// Receives an event from Astarte.
     ///
+    /// When receiving a [`None`] event, the device is disconnect.
+    ///
     /// # Note
     ///
     /// An event can only be received once, so if the client is cloned only one of the clients
     /// instances will receive the message.
-    fn recv(&self) -> impl Future<Output = Result<DeviceEvent, RecvError>> + Send;
+    fn recv(&self) -> impl Future<Output = Option<DeviceEvent>> + Send;
```
lucaato, line 271: this one seems really useful

src/client/mod.rs
```diff
@@ -643,19 +581,17 @@ where
...
         self.send_unset(interface_name, &path).await
     }
 
-    async fn recv(&self) -> Result<DeviceEvent, RecvError> {
-        self.events
-            .recv()
-            .await
-            .map_err(|_| RecvError::Disconnected)?
+    async fn recv(&self) -> Option<DeviceEvent> {
+        self.events.recv().await.ok()
```
lucaato: should we log something like channel closed here ?

src/connection/mod.rs
```diff
@@ -186,32 +187,39 @@ where
...
-                interface: event.interface,
-                path: event.path,
-                data,
-            })?;
+            .await;
+
+        let data = match data {
+            Ok(data) => data,
+            Err(error) => {
+                error!(error = %Report::new(error), "error in received event");
```
lucaato, line 201: I think this raises the question, should we only expose a poll function (to be called in a loop) instead of a `handle_events` so that we can bubble up these errors ?
This way it could be the user to decide whether to log them or do something else.
For example errors of a missing mapping could be used to update the interface version by some clients

src/connection/mod.rs
lucaato:
```suggestion
            Either::Left((send_res, _)) => send_res.inspect(|()| trace!("event sent to clients")),
```

src/transport/mqtt/mod.rs
```diff
@@ -466,28 +486,34 @@ where
         &mut self,
         interfaces: &Interfaces,
         removed: &Interface,
-    ) -> Result<(), Error> {
+    ) -> Result<(), AstarteError> {
```
lucaato, line 489: These are very useful improvements.
It would be also cool to be able to avoid returning a super generic error here.
For example it seems to me that here we are always retunrning `Mqtt`, or `Session`.
Maybe having a smaller enum containing only the error codes used in the the specific user facing function could improve developer experience even further.
For example a `RemoveInterfaceError` that contains only the aforementioned type of errors. There would need to be a new error kind for the pub functions but it seems doable.

lucaato (changes requested):

astarte-device-error/src/lib.rs
joshuachp: The context doesn't require allocations but can reuse and error kind in multiple places while making the error distinct. For example, an io::ErrorKind::NotFound can have a context "while reading interface" and another "interface directory", and this is sufficient in many cases.

While with message you have the possibility to allocate the path used with `path.display().to_string()` and giving more information (but not required).

src/connection/mod.rs
joshuachp: This was what we provided before with the client having the receive function returning a Result.

In most cases those errors should be ignored and just logged which we can do internally to the SDK, while if an error is returned (eg. StoreError) we should actually crash.

For this reason I think it's simpler to handle the error internally to the sdk and log them here.

astarte-device-error/src/lib.rs
joshuachp: Those are sorted in display order

src/error.rs
joshuachp: Display-doc is useful but has the unfortunate downside of showing parameters etc in the documentation. Even if, I initially introduced it, I would prefer to remove it from the projects gradually.

src/transport/mqtt/mod.rs
joshuachp: I think this could be done in another PR. Maybe when we make the mqtt connection feature gated.

But I wouldn't change it here, since it's usefull to return the AstarteError, since one of the advantages is that we can reuse the general error kinds like `Io` or `InterfaceError`.

astarte-device-error/src/lib.rs
lucaato: what about swapping the two names does it make more sense to have a message be a str and the context be a Display Debug instance ?

src/transport/mqtt/mod.rs
lucaato: Okay if you think it's useful we can open an issue to explore it further

astarte-device-error/src/lib.rs
joshuachp: Sure, make sense 👍🏻

lucaato (approved):

PR #653, opened 2026-07-08
ci: fix release not working on master

joshuachp:


lucaato (approved):

.github/workflows/ci.yaml
```diff
@@ -22,6 +22,7 @@ on:
   pull_request:
   push:
     branches:
+      - main
```
harlem88, line 25: Why do we need to add main if it is not part of the repo branches?

harlem88 (approved):

PR #665, opened 2026-07-14
feat(derive)!: add the required object attribute

joshuachp:
Make all the object fields `Option` by default in the FromEvent macro, and add the `#[mapping(required)]` attribute to mark them as required.

This change better reflects the behavior of the Astarte Interface schema.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -238,14 +238,15 @@ mod tests {
         assert_eq!(val.clone().try_into_property(), Ok(prop));
     }
 
-    #[test]
     #[cfg(feature = "derive")]
-    fn should_derive_form_event_obj() {
-        use crate::aggregate::AstarteObject;
-        use crate::{DeviceEvent, FromEvent, Value};
+    mod derive {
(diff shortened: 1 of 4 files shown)


src/event.rs
```diff
@@ -238,108 +238,164 @@ mod tests {
         assert_eq!(val.clone().try_into_property(), Ok(prop));
     }
 
-    #[test]
     #[cfg(feature = "derive")]
-    fn should_derive_form_event_obj() {
-        use crate::aggregate::AstarteObject;
-        use crate::{DeviceEvent, FromEvent, Value};
+    mod derive {
```
lucaato, line 242: is the case of missing required properties already tested ?

lucaato (commented):
should we also modify the examples and the e2e ?

src/event.rs
joshuachp: Propeties cannot have the required attribute since its parsed only in structs (object aggregates)

rgallor (approved):

PR #690, opened 2026-07-29
ci: merge repository

joshuachp:


.github/workflows/ci.yaml
```diff
@@ -24,11 +24,21 @@ on:
     branches:
       - main
       - master
-      - release-*
+      - main
```
rgallor: `main` is inserted twice

rgallor: furthermore, do we really need to use `main` in the sdk? should we have only `master`?

.github/workflows/publish.yaml
```diff
@@ -22,6 +22,7 @@ on:
   workflow_dispatch:
   push:
     branches:
+      - master
```
rgallor: also here `master` is inserted twice

rgallor (changes requested):

.github/workflows/ci.yaml
joshuachp: Yes we want both to have the same CI file everywhere

rgallor (approved):

